Issue #2, opened 2016-02-19
Switch component


nathaniellee:
A toggle that uses the on/off switch visualization made popular by smartphone OSes to reflect its current state. This is meant to be used as a replacement for a native check box element.

Related Jira issue: ANXR-216

PR #3, opened 2016-02-19
#2 -- Added the Switch component.

nathaniellee:


jondlm:
Looks good to me.

src/components/Switch/Switch.less
```diff
@@ -0,0 +1,130 @@
...
+@Switch-container-offset: 3px;
+@Switch-container-width: 44px;
+
+@Switch-glow-offset: 5px;
+@Switch-glow-radius: 9px;
+@Switch-glow-scale: scale(1.556);
+
+@Switch-handle-offset: 6px;
+@Switch-handle-radius: 8px;
+@Switch-handle-scale: scale(1.125);
```
ogupte: is the intention to have these LESS vars be globally available?

jondlm: Oh yeah, there should be in `variables.less`

nathaniellee: Yeah I was talking to Dan about that on Friday. Are we just going to have every component's variables in that file? That could get pretty huge. Is there a problem with having individual files for each component that just get imported into `variables.less`?

alisonailea: If we are moving toward using CSS Modules then every file should be responsible for it's own variables.

src/components/Switch/Switch.jsx
```diff
@@ -0,0 +1,121 @@
+import _ from 'lodash';
+import classNames from 'classnames';
+import React from 'react';
+
+import { bindClassNames } from '../../util/style-helpers';
+
+const generateBoundClassNames = bindClassNames('Switch');
+const {
+	bool,
```
alisonailea: indentation?

src/components/Switch/Switch.jsx
```diff
@@ -0,0 +1,121 @@
...
+				<input
+						{...passThroughs}
+						checked={isSelected}
+						className={generateBoundClassNames('native')}
+						disabled={isDisabled}
+						defaultChecked={isSelected}
+						ref='nativeElement'
+						type='checkbox'
+				/>
+				<span className={generateBoundClassNames('visualization-container')} />
```
alisonailea: This should be `<label>` with a `:before` and `:after`pseudo components in the CSS

src/components/Switch/Switch.less
```diff
@@ -0,0 +1,130 @@
...
+		background-color: @color-default-borderColor;
+		border-radius: @Switch-container-height / 2;
+		height: @Switch-container-height;
+		left: @Switch-container-offset;
+		position: absolute;
+		top: @Switch-container-offset;
+		width: @Switch-container-width;
+	}
+
+	&-visualization-glow {
```
alisonailea: this should be a pseudo component `:before`

src/components/Switch/Switch.less
```diff
@@ -0,0 +1,130 @@
...
+		height: @Switch-glow-radius * 2;
+		left: @Switch-glow-offset;
+		opacity: 0;
+		position: absolute;
+		top: @Switch-glow-offset;
+		transition: @Switch-animation-timing;
+		width: @Switch-glow-radius * 2;
+	}
+
+	&-visualization-handle {
```
alisonailea: this should be a pseudo component `:after`

src/components/Switch/Switch.jsx
nathaniellee: We decided to stick with the named `span` elements since the pseudo element approach added cognitive load when reading the code.

src/components/Switch/Switch.jsx
nathaniellee: Those are tabs in my code so it might be an issue with whatever is rendering the code on the GitHub page.

ogupte:
I approve this PR

src/components/Switch/Switch.less
jondlm: As long as we're careful about keeping and using global variables from `variables.less` that this shouldn't be an issue. I'm cool with keeping component specific variables in their respective files.

src/components/Switch/Switch.jsx
jondlm: Yeah, I noticed that as well. GitHub doesn't show tabs as well as stash does.

src/components/Switch/Switch.jsx
jondlm: Yup. Agreed with @nathaniellee on this one.

jondlm:
:v: My name is Jon and I approve this PR. :v:

PR #8, opened 2016-02-19
Feature/#5 icons + Badges

alisonailea:


jondlm:
We need to fix the git history on this commit. Looks like you merged all the old stuff back in.

PR #14, opened 2016-02-23
Refactor child-component.js and fix bug

### LINKED ISSUE #13: Fix child component bug and improve function names
There is a bug with `getAllAsProps` in the `child-components` file. We should also change some of the names of those functions to be more clear.

jondlm:
There was a bug that was caused by not filtering out `null` values in the old version of `getAllAsProps`, this was fixed by `.filter(x => x)`

### DIFF
--- a/src/util/child-component.js
+++ b/src/util/child-component.js
@@ -90,8 +90,8 @@ export function createChildComponent (definition={}) {
 			getAllAsProps(parentProps) {
 				let component = this;
 
-				var propsFromProps = component.getOwnProps(parentProps);
-				var propsFromChildren = component.getOwnChildrenAsProps(parentProps.children);
+				const propsFromProps = component.getOwnProps(parentProps);
+				const propsFromChildren = component.getOwnChildrenAsProps(parentProps.children);
 
 				return propsFromProps.concat(propsFromChildren);
 			}
--- a/src/util/child-component.js
+++ b/src/util/child-component.js
@@ -18,68 +18,82 @@ export function createChildComponent (definition={}) {
 			/**
 			 * getOwnChildren
 			 *
-			 * Returns all collective children of elements with this component type
+			 * Returns all elements with this component type
 			 *
-			 * @param {Array} elements - usually `this.props.children` but can be any list of elements which
-			 *	                          might have this type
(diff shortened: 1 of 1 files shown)


src/util/child-component.js
```diff
@@ -18,68 +18,82 @@ export function createChildComponent (definition={}) {
...
+					.filter(x => x) // remove falsey values
+					.value();
 
-			getAllAsProps(props, elements) {
-				let component = this;
-				return _.map(_.flatten([_.get(props, definition.propName, [])]), (childProp) => {
+				return _.map(currentProps, (childProp) => {
 					if (_.isPlainObject(childProp)) {
-						return _.assign({}, childProp, {children: childProp.children || childProp.text});
+						return childProp;
```
jondlm: I got rid of the `text` alias cause I thought it was confusing. We can always add it back if needed.

src/util/child-component.js
```diff
@@ -18,68 +18,82 @@ export function createChildComponent (definition={}) {
...
-				let component = this;
-				return _.map(component.getAll(elements), 'props');
-			},
+			getOwnProps(parentProps) {
+				// Extract out the defined `childProps` from the parent class
+				const currentProps = _.chain(parentProps)
+					.get(definition.propName) // grab the prop we care about, e.g. "Child"
+					.thru(x => [x]) // wrap in an array
+					.flatten()
+					.filter(x => x) // remove falsey values
```
jondlm: This is the bug fix line

src/util/child-component.js
```diff
@@ -18,68 +18,82 @@ export function createChildComponent (definition={}) {
...
-			 *	                          might have this type
-			 * @return {Array} `props` objects of elements with this component type
+			 * @param {object} parentProps - usually `this.props` that may contain a prop for the current component type
+			 * @return {array} - an array of normalized props
 			 */
-			getAllProps(elements) {
-				let component = this;
-				return _.map(component.getAll(elements), 'props');
-			},
+			getOwnProps(parentProps) {
```
jondlm: Broke out a new function to make the logic a little clearer

ogupte:
looks good

PR #15, opened 2016-02-23
Add unit tests for state-management functions

ogupte:


src/util/state-management.spec.js
```diff
@@ -0,0 +1,549 @@
...
+	});
+});
+
+describe('#bindReducerToState', () => {
+	it('should bind a single reducer function to a state management interface', () => {
+		let state = {
+			value: null
+		};
+
+		const stateManager = {
```
risingspiral: Can we DRY up these stateManager declarations?

ogupte: I _could_ DRY them up a bit by adding `let state;` up to the shared scope and then have stateManager live with it, while each test assigns `state` a specific value individually. But I feel it takes away from isolation and readability of each test by segregating interdependent data. Generally, I feel tests can be too DRY for that reason.

jondlm: I agree with @ogupte. I think it's acceptable to sacrifice DRYness in favor of isolation and readability of tests.

src/util/state-management.spec.js
```diff
@@ -0,0 +1,549 @@
+import assert from 'assert';
+import _ from 'lodash';
+import sinon from 'sinon';
+import React from 'react';
+import { mount } from 'enzyme';
```
jondlm: If you're using `mount` it means you need the DOM. Make sure you use the `describeWithDOM` helper to make it clear that those tests require the DOM.

jondlm:
Approved

Issue #18, opened 2016-02-23
Banner Component


danhahn:
Build a banner component 

Related to jira ticket ANXR-194

- [x] create component
- [ ] add tests

ogupte:
more tests are needed for the various Banners

PR #17, opened 2016-02-23
Feature/ icons

alisonailea:
Reopening for Travis build

PR #19, opened 2016-02-23
#18 - Added Banner component

danhahn:


src/components/Banner/Banner.jsx
```diff
@@ -0,0 +1,153 @@
...
+	'danger': '<DangerIcon/>',
+	'info': '<InfoIcon/>',
+	'warning': '<WarningIcon/>',
+	'primary': null,
+	'default': null
+};
+
+/**
+ *
+ * {"categories": ["controls", "Banners"]}
```
jondlm: Categories should be lower case. The code that reads them passes them through the [`_.startCase`](https://lodash.com/docs#startCase) function to format them.

src/components/Banner/Banner.jsx
```diff
@@ -0,0 +1,153 @@
...
+			icon: null,
+			isCloseable: true,
+			isFullWidth: false,
+			kind: 'default',
+			onClick: _.noop
+		};
+	},
+
+	handleClick() {
+		this.props.onClick();
```
jondlm: Since we aren't doing anything special for `onClick` we should just pass `this.props.onClick` directly without a handler.

src/components/Banner/Banner.jsx
```diff
@@ -0,0 +1,153 @@
...
+			<section
+				className={classNames(className, scopedClasses)}
+				onClick={this.handleClick}
+				{...others}
+				ref='Banner' >
+				{displayedIcon ? <span className={boundClassNames('icon')}>{displayedIcon}</span> : null}
+				<span className={boundClassNames('content')}>
+					{children}
+				</span>
+				{isCloseable ? <span className={boundClassNames('close')} onClick={this.handleClick}>{ String.fromCharCode(215) } </span> : null}
```
jondlm: Should we have another prop called `onClose` instead of using onClick for two different things?

src/components/Banner/Banner.less
```diff
@@ -0,0 +1,90 @@
+@banner: lucid-Banner;
```
jondlm: Either we should make these "name" variables more specific or we should do away with them.

I'm thinking either something more like `Banner-name` if we need to use a variable.

src/components/Banner/Banner.jsx
```diff
@@ -0,0 +1,153 @@
...
+	oneOfType,
+	oneOf,
+	node,
+	arrayOf,
+	func,
+	element
+} = React.PropTypes;
+
+const defaultIcons = {
+	'success': '<SuccessIcon/>',
```
jondlm: I'm guessing this is just a placeholder until #17 lands?

src/components/Banner/Banner.jsx
```diff
@@ -0,0 +1,153 @@
...
+			'danger': kind === 'danger',
+			'info': kind === 'info',
+			'small': size === 'small',
+		});
+
+		return (
+			<section
+				className={classNames(className, scopedClasses)}
+				onClick={this.handleClick}
+				{...others}
```
jondlm: Call these `passThroughs` instead of `others`. It's a slightly better name we've been going with.

Also, move this spread operator up two lines so the stuff you wrote can overwrite what a user passes in.

src/components/Banner/Banner.less
```diff
@@ -0,0 +1,90 @@
...
+		padding-right: 10px;
+	}
+
+	&.@{banner}-full-width {
+		border-radius: 0;
+	}
+
+	&-has-icon {
+		& > .@{banner}-content {
+			padding: @size-banner-spacing-topBottom + 2px 0;
```
jondlm: For component specific variables we landed on `@Banner-size-spacing-topBottom`. See the CONTRIBUTING doc for a couple more examples. And feel free to propose changes to that doc if you feel strongly about anything.

src/components/Banner/Banner.less
```diff
@@ -0,0 +1,90 @@
...
+		}
+	}
+
+	&-content {
+		flex: 1;
+		padding: @size-banner-spacing-topBottom 0;
+		line-height: 1em;
+		& > p {
```
jondlm: Do we expect consumers to be putting `p` tags as content? We should probably call that out somewhere if that's true.

src/styles/variables.less
```diff
@@ -26,36 +38,41 @@
...
 @color-default-borderColor: #c5c5c5;
-@color-default-backgroundColor: transparent;
+@color-default-backgroundColor: #ededed;
 @color-default-gradientStartColor: #f3f3f3;
 @color-default-gradientEndColor: #e2e2e2;
 
 // primary
 @color-primary-borderColor: @color-primary;
-@color-primary-backgroundColor: transparent;
+@color-primary-borderColorLite: fade(@color-primary, 40%);
```
jondlm: Cool, nice additions to the colors.

src/styles/variables.less
```diff
@@ -11,10 +18,15 @@
 
 @size-generic-boxShadow: rgba(0,0,0,.3);
 
+// banner sizes
+@size-banner-spacing-rightLeft: 10px;
```
jondlm: We're putting component specific variables at the top of the respective component less files now. So these should be moved to `Banner.less`.

src/components/Banner/Banner.less
jondlm: Whoops, looks like those changes to CONTRIBUTING aren't in `master` yet. They are landing in #25

src/components/Banner/Banner.jsx
danhahn: Yes icons had not been merge in as the time of the PR

src/components/Banner/Banner.jsx
```diff
@@ -0,0 +1,154 @@
...
+		 */
+		icon: element,
+		/**
+		 * set this to `true` if you want to have a `x` close icon.
+		 */
+		isCloseable: bool,
+		/**
+		 * set this to `true` if you want to remove rounded corners.
+		 */
+		isFullWidth: bool,
```
nathaniellee: Does this comment accurately portray the purpose of this prop? If so, the prop should probably be renamed. If not, the comment should probably be more accurate. :)

src/components/Banner/Banner.jsx
```diff
@@ -0,0 +1,154 @@
...
+			'warning',
+			'danger',
+			'info'
+		]),
+		/**
+		 * size variations of the Banner
+		 */
+		size: oneOf([
+			'small'
+		]),
```
nathaniellee: If the `size` prop can only have one possible value should it be a prop?

src/components/Banner/Banner.jsx
```diff
@@ -0,0 +1,154 @@
...
+			icon: null,
+			isCloseable: true,
+			isFullWidth: false,
+			kind: 'default',
+			onClick: _.noop
+		};
+	},
+
+	render() {
+		let {
```
nathaniellee: `const`

src/components/Banner/Banner.jsx
```diff
@@ -0,0 +1,154 @@
...
+		if(icon) {
+			displayedIcon = icon;
+		}
+		else if(hasIcon) {
+			displayedIcon = defaultIcons[kind];
+		}
+
+		//if icon than hasicon needs to be ture
+
+		let scopedClasses = boundClassNames('~', {
```
nathaniellee: `const`

src/components/Banner/Banner.jsx
```diff
@@ -0,0 +1,154 @@
...
+			'warning': kind === 'warning',
+			'danger': kind === 'danger',
+			'info': kind === 'info',
+			'small': size === 'small',
+		});
+
+		return (
+			<section
+				className={classNames(className, scopedClasses)}
+				onClick={this.handleClick}
```
nathaniellee: I don't see a `handleClick` method defined anywhere on the component.

src/components/Banner/Banner.jsx
```diff
@@ -0,0 +1,154 @@
...
+				{...passThroughs}
+				ref='Banner' >
```
nathaniellee: I don't see this `ref` being used anywhere.

src/styles/variables.less
```diff
@@ -1,3 +1,10 @@
+// Colors
+
+@black: black;
```
jondlm: Move this in to the existing colors section and call it `@color-black`

src/components/Banner/Banner.less
```diff
@@ -0,0 +1,90 @@
...
+.@{banner} {
+	.box-sizing();
+	border: 1px solid @color-default-borderColor;
+	background-color: @color-default-backgroundColor;
+	font: @font;
+	border-radius: @Banner-size-borderRadiaus;
+	padding: 0 @Banner-size-spacing-rightLeft;
+	display: flex;
+
+	&.@{banner}-has-close {
```
jondlm: Couldn't this just be `&-has-close`? Or am I reading this wrong?

src/components/Banner/Banner.less
```diff
@@ -0,0 +1,90 @@
...
+		padding-right: 10px;
+	}
+
+	&.@{banner}-full-width {
```
jondlm: [Same](https://github.com/appnexus/lucid/pull/19/files#r54446495)

src/components/Banner/Banner.jsx
ogupte: i think it should also have a `normal` size, but if you're just testing the existence of this prop being passed it should really be a `bool` in the patter of `isSmall` to be consistent with other component props

src/components/Banner/Banner.jsx
```diff
@@ -0,0 +1,154 @@
...
+			...passThroughs
+		} = this.props;
+
+		let displayedIcon = null;
+
+		if(icon) {
+			displayedIcon = icon;
+		}
+		else if(hasIcon) {
+			displayedIcon = defaultIcons[kind];
```
ogupte: is it a good idea to use `_.get(defaultIcons, kind, 'default')` so that there's a fallback to `default` if not found?

src/components/Banner/examples/basic.jsx
```diff
@@ -0,0 +1,131 @@
...
+		string
+	}
+} = React;
+
+
+const InfoIcon = React.createClass({
+	propTypes: {
+		className: string,
+		color: string,
+		size: number,
```
ogupte: maybe we need to name this size prop something different since other components use `size: oneOf(['small', 'normal', 'large']) while this one expects a number. i think we should be more consistent.

src/components/Banner/examples/basic.jsx
```diff
@@ -0,0 +1,131 @@
...
+		viewBox: string
+	},
+
+	getDefaultProps() {
+		return {
+			className: '',
```
ogupte: should default to `null` rather than a null string

src/components/Banner/examples/basic.jsx
```diff
@@ -0,0 +1,131 @@
...
+		const {
+			className,
+			color,
+			size,
+			viewBox
+		} = this.props;
+
+		return (
+			<svg
+					className={classNames('ArPlusIcon', className)}
```
ogupte: `ArPlusIcon` should be renamed

src/components/Banner/examples/basic.jsx
```diff
@@ -0,0 +1,131 @@
...
+
+
+export default React.createClass({
+	render() {
+		return (
+			<div>
+				<div>
+					<Banner>Default</Banner>
+					<Banner isFullWidth={true}>Default full width</Banner>
+					<Banner isCloseable={false}>Default -- No Close {String.fromCharCode(215)}</Banner>
```
ogupte: i think we can use the char `'×'` since it's utf8, or if you still want to use a code, i would rather use it in a string literal `'\u00d7'`

src/components/Banner/Banner.jsx
danhahn: change to `isSmall` which is a `bool`

src/components/Banner/Banner.jsx
danhahn: was not needed so I removed it

src/components/Banner/Banner.jsx
danhahn: removed

src/styles/variables.less
danhahn: Fixed

src/components/Banner/Banner.less
danhahn: no because we need the `lucid-Bannder` class name.  Since two classes are being added to the elements each having the `lucid-Banner` name.

src/components/Banner/examples/basic.jsx
danhahn: good point.  this is just a place holder to see what it would look like if a user would put a custom icon.

src/components/Banner/Banner.jsx
danhahn: Added the icons

src/components/Banner/Banner.jsx
danhahn: changed to `hasRoundedCorners` that is set to true by default.

src/components/Banner/Banner.jsx
danhahn: flexed

src/components/Banner/Banner.jsx
danhahn: fixed

src/components/Banner/examples/basic.jsx
jondlm: +1

src/components/Banner/examples/basic.jsx
jondlm: Yeah, this shouldn't pass the `common` tests

jondlm:
Feel free to clone/reopen this PR if the Travis check doesn't work

halseyhemingway:
UX Approved. I can't figure out how to add a label since I'm not a contributor

danhahn:
@jondlm the issue was with a conflict merging in master.

danhahn:
Adding label for @halseyhemingway

Issue #7, opened 2016-02-19
RadioButton component


nathaniellee:
A toggle meant to be used as a replacement for a native radio button element.

Related Jira issue: ANXR-280

PR #22, opened 2016-02-23
#7 - Added RadioButton component

nathaniellee:


src/components/RadioButton/RadioButton.jsx
```diff
@@ -0,0 +1,125 @@
+import _ from 'lodash';
+import classNames from 'classnames';
+import React from 'react';
+
+import { bindClassNames } from '../../util/style-helpers';
+
+const generateBoundClassNames = bindClassNames('RadioButton');
```
jondlm: We should really stick with one name for these.

src/components/RadioButton/RadioButton.jsx
```diff
@@ -0,0 +1,125 @@
...
+		isDisabled: bool,
+
+		/**
+		 * Indicates that the component is in the "selected" state when true
+		 * and in the "unselected" state when false.
+		 */
+		isSelected: bool,
+
+		/**
+		 * Called when the user clicks on the component or when they press the
```
jondlm: You wrap at 80 characters, it's... so... beautiful... :joy: Almost no ones this anymore hah.

src/components/RadioButton/RadioButton.spec.jsx
```diff
@@ -0,0 +1,112 @@
...
+
+	const booleanValues = [true, false];
+
+	describe('props', () => {
+		describe('isDisabled', () => {
+			it('sets the `disabled` attribute of the native radio button element.', () => {
+				_.forEach(booleanValues, (testValue) => {
+					const wrapper = shallow(<RadioButton isDisabled={testValue} />);
+
+					assert.equal(wrapper.find('input[type="radio"]').prop('disabled'), testValue);
```
jondlm: Glad to see this selector works!

src/components/RadioButton/RadioButton.spec.jsx
```diff
@@ -0,0 +1,112 @@
...
+		describe('onSelect', () => {
+			it('defaults to the Lodash `noop` method.', () => {
+				const wrapper = mount(<RadioButton />);
+
+				assert.equal(wrapper.prop('onSelect'), _.noop);
+			});
+		});
+
+		describe('pass throughs', () => {
+			it('passes through all props not defined in `propTypes` to the native input.', () => {
```
jondlm: I was wondering if we could move this kind of test into a common test for "control" type components.

src/components/RadioButton/examples/interactive.jsx
```diff
@@ -0,0 +1,34 @@
...
+		this.setState(_.assign({}, this.state, {
+			isSelected
+		}));
+	},
+
+	render() {
+		return (
+			<ul>
+				<li>
+					<label>Enabled</label>
```
jondlm: Once we have a pattern for putting labels with our control components, we should refactor the examples to use that pattern.

jondlm:
Noice. Approved.

src/components/RadioButton/RadioButton.jsx
nathaniellee: I do wrap at 80 characters. Some of my co-workers get annoyed when pairing with me on my screen because I have Sublime's demarcation line at 80 characters all the way down the window. :)

src/components/RadioButton/RadioButton.jsx
nathaniellee: :P

src/components/RadioButton/RadioButton.jsx
nathaniellee: Never mind. I misread your comment as "You _should_ wrap at 80 characters." LOL

src/components/RadioButton/examples/interactive.jsx
nathaniellee: Yeah that's how my question around how our labels should work the other day came up -- Dave and Ashley were requesting that labels be included in the examples.

src/components/RadioButton/RadioButton.spec.jsx
nathaniellee: I had thought about it: we would need to be able to pass in the names of the component's actual props so we know what _shouldn't_ be passed through and we would also need to be able to pass through the selector somehow since in some cases the native is the root of the render whereas in others it's a child of the root (like here).

One point of trouble might be if for some reason on some components we need to pass through some of those defined props down to the native. We haven't done enough of these to know if it's an edge case or not.

src/components/RadioButton/RadioButton.jsx
jondlm: Even `boundClassNames` is a little long for my taste since it's so ubiquitous.

src/components/RadioButton/RadioButton.jsx
```diff
@@ -0,0 +1,125 @@
...
+		const {
+			className,
+			isDisabled,
+			isSelected,
+			style,
+			...passThroughs
+		} = this.props;
+
+		return (
+			<div
```
alisonailea: since this may not always be used as a block level element we should change this to a <span>

src/components/RadioButton/RadioButton.jsx
nathaniellee: `Verbose and explicit` > `concise and unclear` ;)

Actually `boundClassNames` doesn't seem too ambiguous to me, but it breaks the general rule of thumb that a function/method should be named as an action verb.

But if the team feels strongly enough that `boundClassNames` is fine I'll begrudgingly concede. ;)

src/components/RadioButton/RadioButton.jsx
nathaniellee: We should briefly discuss this as I originally was going with `span` elements as the root for these components and Dan said that we should use `div` elements instead. My thought was what you said: that it's not likely a block-level element, but I conceded to his opinion since he was more experienced as a semantic markup type developer. :)

PR #33, opened 2016-02-25
#32 - Checkboxes

jondlm:


jondlm:
Tested on Chrome/Firefox/Safari/IE11

alisonailea:
approve

src/components/Checkbox/Checkbox.jsx
```diff
@@ -0,0 +1,125 @@
...
+		 */
+		isSelected: bool,
+
+		/**
+		 * Called when the user clicks on the component or when they press the
+		 * space key while the component is in focus, and only called when the
+		 * component is in the unselected state. The first parameter is true,
+		 * and the second parameter is the synthetic event object generated by
+		 * React.
+		 */
```
nathaniellee: "The first parameter is true" -- This is only applicable to the `RadioButton` component since that one cannot be turned "off" via user interaction.

The analogous comment in the `Switch` component is:

```
		/**
		 * Called when the user clicks on the component or when they press the
		 * space key while the component is in focus: the first parameter is a
		 * boolean value set to the opposite of the isSelected prop's value
		 * and the second parameter is the synthetic event object generated by
		 * React.
		 */
```

nathaniellee:
Looks good! Just the one edit to the comments for the `onSelect` prop. :)

src/components/Checkbox/Checkbox.jsx
jondlm: Ahh good catch, thanks for seeing my copy paste mistake heh.

[ ] fix this

ogupte:
a test failed because CheckBox is never exported in `index.js`

jondlm:
Classic ;) I added the test but forgot the fix the thing I added it for. Shouldn't be an issue moving forward now.

Issue #36, opened 2016-02-25
Improve docs page


jondlm:
- [x] add sidebar for navigation
- [x] add code collapsing
- [x] differentiate "composite" components. I believe docgen already has `Component.composes` which suits our need.
- [x] support markdown in the description and put in on the page
- [x] support markdown in props descriptions
- [x] sort props alphabetically
- [x] show default prop values
- [x] add support for ordering examples
- [x] fix nav scroll overflow

jondlm:
Not quite finished with all the tasks on this one.

PR #38, opened 2016-02-27
#36 - Docs improvements

jondlm:
- differentiate "composite" components. I believe docgen already has Component.composes which suits our need.
- support markdown in the description and put in on the page
- sort props alphabetically


jondlm:
Also added markdown parsing to prop comments

src/docs/index.less
```diff
@@ -37,6 +37,11 @@ html {
 		justify-content: space-between;
 		align-items: center;
 	}
+
+	&-props-table {
+		p { margin: 5px 0; }
+		td { padding: 2px 7px; }
```
ogupte: probably format these rule sets to match the format of other css rules sets in the file.

jondlm: Will do!

Issue #48, opened 2016-03-01
Add color palette


danhahn:
Update the color palette to match designs

jondlm:
Accidentally closed this. I thought the PR was merged.

PR #49, opened 2016-03-01
Add color palette #48

danhahn:
Standardizing colors


src/styles/variables.less
```diff
@@ -25,12 +40,8 @@
 @color-warning: #feb209;
 @color-danger: #f7403a;
 
-@color-white: #fff;
-
-@color-gray-5: fade(#000, 5%);
-@color-gray-10: fade(#000, 10%);
-@color-gray-25: fade(#000, 25%);
-@color-gray-30: fade(#000, 30%);
+@color-borderColor: @color-gray;
```
ogupte: when would you use `@color-borderColor` verses `@color-default-borderColor`?

danhahn: `default` is related to the default button coloring.  I would think that we would use `@color-borderColor` anytime we need to define a border we would set the color using this variable

src/styles/variables.less
```diff
@@ -25,12 +40,8 @@
...
+@color-backgroundColor: @color-ligthGray;
 
 // default
 @color-default-borderColor: #c5c5c5;
```
danhahn: no it is its own color defined for the user in a `<Button>` or `<Banner>`

src/styles/variables.less
```diff
@@ -18,6 +18,21 @@
...
+@color-gray-5: fade(@color-black, 5%);
+@color-gray-10: fade(@color-black, 10%);
+@color-gray-25: fade(@color-black, 25%);
+@color-gray-30: fade(@color-black, 30%);
+
+// basic colors
+@color-ligthGray: #f4f4f4;
+@color-gray: #e3e3e3;
+@color-mediumGray: #c5c5c5;
+@color-darkGray: #333333;
```
danhahn: @jondlm added them for one of the components.  I think there is room for consolidation around this

jondlm: I agree we should consolidate these. Also, I'm assuming all those gray colors could be built using a `fade`.

src/styles/variables.less
```diff
@@ -18,6 +18,21 @@
 @size-button-borderRadius: @size-borderRadius;
 
 // colors
+@color-white: #fff;
+@color-black: black;
```
nathaniellee: Why does `@color-white` use the hex `#fff` value while `@color-black` uses the color name `black`?

danhahn: they should both use hex colors

src/styles/variables.less
```diff
@@ -18,6 +18,21 @@
...
+
+@color-gray-5: fade(@color-black, 5%);
```
jondlm: I vote that we use only one set of grays. It's weird to be using two sets.

jondlm:
I think it makes more sense to move nearly all the logic in the example in to the `ColorPalette` component since it's really specific to our library. 

cc @ogupte @nathaniellee

src/styles/variables.less
nathaniellee: I'm not sure I understand why we couldn't use `@color-mediumGray` as the value of `@color-default-borderColor` since `@color-mediumGray` is `#c5c5c5`.

src/components/Button/Button.less
```diff
@@ -8,11 +8,7 @@
 	display: inline-block;
 	outline: none;
 	border: 1px solid @color-default-borderColor;
-	border-radius: @size-button-borderRadius;
```
jondlm: Did you mean to remove button styles?

src/styles/variables.less
jondlm: Not sure I see the distinction either.

src/docs/index.less
```diff
@@ -77,3 +78,222 @@ html {
 		}
 	}
 }
+
```
ogupte: should these styles perhaps live in their own less file at `src/docs/containers/colors.less` maybe?

jondlm: Probably, I've just been throwing everything in index until it grows too big.

jondlm: Not needed for this PR I think though

PR #51, opened 2016-03-01
Added RadioGroup component.

nathaniellee:


src/util/random.js
```diff
@@ -0,0 +1,3 @@
+export default function () {
+	return Math.random().toString(16);
```
ogupte: we should use: `Math.random().toString(16).substr(2)` since we want a random hex string rather than '0.{hex}'.

src/components/RadioGroup/RadioGroup.spec.jsx
```diff
@@ -0,0 +1,81 @@
...
+		describe('onSelect', () => {
+			it('defaults to the Lodash `noop` method.', () => {
+				const wrapper = shallow(<RadioGroup />);
+
+				assert.equal(wrapper.prop('onSelect'), _.noop);
+			});
+		});
+
+		describe('pass throughs', () => {
+			it('passes through all props not defined in `propTypes` to the root element.', () => {
```
ogupte: i'm not sure how valuable this test is

src/components/RadioGroup/RadioGroup.spec.jsx
```diff
@@ -0,0 +1,81 @@
...
+				const name = wrapper.first().prop('name');
+
+				_.forEach(wrapper.find('RadioButton').nodes, (node) => {
+					assert.equal(node.props.name, name);
+				});
+			});
+		});
+
+		describe('onSelect', () => {
+			it('defaults to the Lodash `noop` method.', () => {
```
ogupte: i don't think this test is very valuable, instead maybe pass in a sinon.spy via `onSelect`, then do something, then assert that it has been called with the correct arguments.

src/components/RadioGroup/RadioGroup.spec.jsx
```diff
@@ -0,0 +1,81 @@
...
+						<RadioGroup.RadioButton />
+					</RadioGroup>
+				);
+
+				_.forEach(wrapper.find('RadioButton').nodes, (node) => {
+					assert.equal(node.props.name, name);
+				});
+			});
+
+			it('defaults to a randomly generated string that is passed along to the children.', () => {
```
ogupte: the test text can leave out the random part since it's not checking for randomness.

src/components/RadioGroup/RadioGroup.spec.jsx
```diff
@@ -0,0 +1,81 @@
+import _ from 'lodash';
+import React from 'react';
+import assert from 'assert';
+// import sinon from 'sinon';
```
ogupte: commented out import

src/components/RadioGroup/RadioGroup.jsx
```diff
@@ -0,0 +1,143 @@
...
+			selectedIndex,
+			...passThroughs
+		} = this.props;
+
+		const radioButtonChildProps = RadioGroup.RadioButton.findInAllAsProps(this.props);
+		const selectedIndexFromChildren = _.findLastIndex(radioButtonChildProps, {
+			isSelected: true
+		});
+
+		// If there are any `RadioGroup.RadioButton` children with `isSelecte`
```
ogupte: `isSelecte` is misspelled :P

src/components/RadioGroup/RadioGroup.jsx
```diff
@@ -0,0 +1,143 @@
...
+		 * selected. The index of the last `RadioGroup.RadioButton` child with
+		 * `isSelected` equal to true takes precedence over this prop.
+		 */
+		selectedIndex: number
+	},
+
+	getDefaultProps() {
+		return {
+			children: null,
+			className: null,
```
ogupte: null default props should be ommitted

src/components/RadioGroup/RadioGroup.jsx
```diff
@@ -0,0 +1,143 @@
...
+	reducers,
+
+	propTypes: {
+		/**
+		 * Should be instances of `RadioGroup.RadioButton` which supports the
+		 * same props as `RadioButton`.
+		 */
+		children: oneOfType([
+			arrayOf(node),
+			node
```
ogupte: just `node` is fine since an array of `node`s is itself a `node`

src/util/random.js
nathaniellee: Good catch. I accidentally left off the call to `substr`.

src/components/RadioGroup/RadioGroup.spec.jsx
nathaniellee: It seems to me that if we're explicitly testing that we're passing through styles then we should be testing that we're passing through other props as well. That sometimes we're passing through to root while other times we're passing through to a native control makes it even more pertinent IMO.

src/components/RadioGroup/RadioGroup.spec.jsx
nathaniellee: My test suite is incomplete here. I tend to test passed in event handlers in more "functional" tests. I always test for default values in my "props" section which I think is important since the default value is a key part of the API definition, no?

src/components/RadioGroup/RadioGroup.spec.jsx
nathaniellee: Makes sense. :)

src/components/RadioGroup/RadioGroup.spec.jsx
nathaniellee: It'll be uncommented out by the time this spec is finished. :)

src/components/RadioGroup/RadioGroup.jsx
nathaniellee: Ack!

src/components/RadioGroup/RadioGroup.spec.jsx
ogupte: it seems like something that should be covered in the generic tests

src/components/RadioGroup/RadioGroup.spec.jsx
ogupte: testing that default props work seems out of scope, since it's safe to assume that React handles default props correctly, imo

src/components/RadioGroup/RadioGroup.jsx
nathaniellee: I don't think that's _technically_ true insofar as what React actually produces in its renders.

```
<RadioGroup name={name}>
	<RadioGroup.RadioButton />
	<RadioGroup.RadioButton />
	<RadioGroup.RadioButton />
</RadioGroup>
```

This results in a `RadioGroup` node that has three `RadioGroup.RadioButton` nodes stored in its `nodes` array.

It's not possible to just render the following:

```
<RadioGroup.RadioButton />
<RadioGroup.RadioButton />
<RadioGroup.RadioButton />
```

But if it were it would result not in a single node entity but rather an array of them.

With all that said, though, the documentation for `React.PropTypes` _does_ note that when we use `React.PropTypes.node` that check also includes arrays of the represented types...

...so, yeah, I'll simplify that statement. :)

src/components/RadioGroup/RadioGroup.spec.jsx
nathaniellee: I agree except that the common, generic tests are invariable whereas this test is going to be different depending on whether the component in question uses a native control as its root or as a child of its root.

src/components/RadioGroup/RadioGroup.spec.jsx
nathaniellee: The intent isn't to test that React itself is passing default props correctly but that in the case of any given component if a specific prop isn't provided by the consumer then the component utilizes the proper default value.

Theoretically, if the expectation of a component is that when `foo` is not defined it has the value `true` which causes some particular part of markup to be included in the rendered output and someone unintentionally changes the default value to `false`, there should be a unit test that breaks to highlight this situation, no?

src/components/RadioGroup/RadioGroup.jsx
```diff
@@ -0,0 +1,136 @@
...
+	},
+
+	reducers,
+
+	propTypes: {
+		/**
+		 * Should be instances of `RadioGroup.RadioButton` which supports the
+		 * same props as `RadioButton`.
+		 */
+		children: node,
```
jondlm: We should standardize this across all our components. I've been using `any`, which I'm not sure is good. Is `node` the best thing for children?

src/components/RadioGroup/RadioGroup.reducers.js
```diff
@@ -0,0 +1,8 @@
+export default {
```
jondlm: This is exactly the same reducer I wrote for `Tabs` :+1:

src/components/RadioGroup/RadioGroup.jsx
```diff
@@ -0,0 +1,136 @@
...
+
+		const radioButtonChildProps = RadioGroup.RadioButton.findInAllAsProps(this.props);
+		const selectedIndexFromChildren = _.findLastIndex(radioButtonChildProps, {
+			isSelected: true
+		});
+
+		// If there are any `RadioGroup.RadioButton` children with `isSelected`
+		// equal to true, then the index of the last one should override the
+		// value of the `selectedIndex` prop.
+		const actualSelectedIndex = selectedIndexFromChildren !== -1 ?
```
jondlm: I really like how you handled `selectedIndex` vs child `isSelected`. I've fixed up Tabs to align with this.

src/util/random.js
```diff
@@ -0,0 +1,3 @@
+export default function () {
+	return Math.random().toString(16).substr(2);
```
jondlm: We should probably use this in `Portal` as well

jondlm:
Behaves as expected in Chrome/Firefox/Safari/IE11. We just need to fix that styling bug in the underlying `RadioButton`

src/components/RadioGroup/RadioGroup.jsx
nathaniellee: I think it should definitely be `node` which represents any "renderable" type (strings, numbers, elements or an array of those types). Anything else isn't renderable so wouldn't be able to be passed in as a child, anyway.

Issue #56, opened 2016-03-02
generic tests to check `displayName` and support required props


ogupte:
some components have requried props or children and generic-tests needs a way to pass those in. also generic-tests uses `displayName` all over the place but never tests that it exists in the first place, so it's confusing when an unrelated test breaks.

PR #57, opened 2016-03-02
#56: common tests accept getDefaultProps function and check `displayName`

ogupte:
see #56


jondlm:
Force merged this one since travis wasn't reporting the success back for some reason.

Issue #58, opened 2016-03-02
Error when running 'npm run dev'


mute:

mute:
Hey there,

I'm trying to run the project locally and getting the following error from react-router:

![screen shot 2016-03-02 at 2 34 21 pm](https://cloud.githubusercontent.com/assets/730349/13478040/00d4cdde-e084-11e5-82f2-fc6af92f26f8.png)

Looks like it might be a dependency version issue?

jondlm:
Fixed by #59

jondlm:
Looks like we're still getting an error in the console about using a deprecated approach. I found the correct solution: https://github.com/reactjs/react-router/blob/master/upgrade-guides/v2.0.0.md#using-custom-histories

I'm fixing it now.

PR #59, opened 2016-03-02
use reaact-router 2.0 browserHistory

jhsu:
react-router 2.0 uses browserHistory 
https://github.com/reactjs/react-router/blob/master/upgrade-guides/v2.0.0.md

addresses #58

- [x] use hash history


mute:
:+1: Grabbed this branch and it fixed the issue I was seeing locally.

jhsu:
updated this to create a custom hash history that doesn't use queryKey

jondlm:
Awesome, thanks for the fix!

Issue #60, opened 2016-03-03
docs need to support example order and nav scroll overflow


ogupte:

ogupte:
closing this issue since it's duplicated in #36

PR #61, opened 2016-03-03
Docs: sidebar scrollable and example names ignore numbers

ogupte:
make sidebar scrollable and example names excludes leading numbers to establish render order.

addresses #60

### DIFF
--- a/src/docs/index.jsx
+++ b/src/docs/index.jsx
@@ -2,6 +2,7 @@
 require('./index.less');
 require('../index.less');
 
+import path from 'path';
 import _ from 'lodash';
 import React from 'react';
 import { render } from 'react-dom';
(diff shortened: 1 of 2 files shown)


src/docs/index.less
```diff
@@ -62,6 +62,7 @@ html {
 		width: @App-sidebar-width;
 		padding: 15px;
 		padding-left: 0;
+		overflow: auto;
```
jondlm: Yep, needed this :)

src/docs/index.jsx
```diff
@@ -33,13 +34,22 @@ const docgenMap = _.mapValues(docgenMapRaw, (value, componentName) => {
...
+ * file extensions then capitalizes each word and separates each with a space.
+ */
+function getExampleTitleFromFilename (filename) {
+    const words = _.words(_.startCase(path.basename(filename, '.jsx')));
+    return (/^\d+$/.test(_.head(words)) ? _.tail(words) : words).join(' ');
+}
+
 const examplesByComponent = _.chain(reqExamples.keys())
 	.map((path) => {
 		const items = path.split('/').reverse(); // e.g. ['default.jsx', 'examples', 'Button', '.']
```
jondlm: Looks like we're using `path` in two different ways now, should rename this one

ogupte: ah missed that

PR #64, opened 2016-03-03
Tabs component

### LINKED ISSUE #45: Tabs component
Create a new component `Tabs`.

jondlm:
Fixes #45 by adding `Tabs` component with examples and tests.


src/styles/variables.less
```diff
@@ -44,6 +44,10 @@
 @color-gray-25: fade(#000, 25%);
 @color-gray-30: fade(#000, 30%);
 
+// Shadows
+
+@shadow-inset: 0 10px 10px -10px @color-gray-30 inset;
```
ogupte: question: does it make sense to name the variable `@box-shadow-inset` to be consistent with the css property?

src/components/Tabs/Tabs.jsx
```diff
@@ -0,0 +1,127 @@
...
+		/**
+		 * Styles that are passed through to the root container.
+		 */
+		style: any,
+
+		/**
+		 * While you can pass tab content through this prop, it's better to use the
+		 * child component instead. See the examples for details.
+		 */
+		Tab: any.isRequired,
```
ogupte: having `isRequired` will give a warning if you decide to just pass thru as children, right?

src/components/Tabs/Tabs.jsx
```diff
@@ -0,0 +1,127 @@
...
+
+		/**
+		 * `Title` can be a prop on `Tabs.Tab` or a child of `Tabs.Tab`. See the
+		 * examples for details.
+		 */
+		Title: any,
```
ogupte: createLucidComponentDefinition already will `assign` the `childProps` proptypes to the component `propTypes`, so having them here is redundant since they are overridden.

ogupte: we'll need to make our docs parse and print out `childProps`

src/components/Tabs/Tabs.jsx
```diff
@@ -0,0 +1,127 @@
...
+		/**
+		 * If `true` then the active tab will appear open on the bottom.
+		 */
+		isOpen: bool,
+	},
+
+	getDefaultProps() {
+		return {
+			style: null,
+			className: null,
```
ogupte: should we make it a practice to assign `null` in default props? i'm not sure if it's useful

src/components/Tabs/Tabs.jsx
jondlm: Yeah that's a good point

- [ ] remove `isRequired` from `Tab`

src/components/Tabs/Tabs.jsx
jondlm: Ahh okay, this was just a work around then. I can remove them until we solve the docgen issue.

src/components/Tabs/Tabs.jsx
jondlm: If it's not needed then I'll remove them. It was just habit really.

src/components/Tabs/Tabs.spec.jsx
```diff
@@ -0,0 +1,120 @@
...
+					<Tabs.Tab Title='Lollipop'>Yuck</Tabs.Tab>
+					<Tabs.Tab Title='Slurpee'>Yum</Tabs.Tab>
+				</Tabs>
+			);
+
+			assert.equal(wrapper.find('.lucid-Tabs-bar-item-is-active-and-open').length, 0);
+		});
+	});
+
+	describe('reducers', () => {
```
ogupte: should reducer tests live in their own `.spec` file?

src/styles/variables.less
jondlm: I was thinking of `shadow` as another top level `type`. Maybe it should be `shadow-boxShadowInset` or something? Just trying to stick to a convention we lay out in [contrib](https://github.com/appnexus/lucid/blob/master/CONTRIBUTING.md)

src/components/Tabs/Tabs.spec.jsx
```diff
@@ -0,0 +1,120 @@
...
+
+		it('isOpen', () => {
+			const wrapper = shallow(
+				<Tabs isOpen={false} selectedIndex={0}>
+					<Tabs.Tab Title='Lollipop'>Yuck</Tabs.Tab>
+					<Tabs.Tab Title='Slurpee'>Yum</Tabs.Tab>
+				</Tabs>
+			);
+
+			assert.equal(wrapper.find('.lucid-Tabs-bar-item-is-active-and-open').length, 0);
```
ogupte: go big or go home

src/components/Tabs/Tabs.spec.jsx
jondlm: I can see the value in that.

src/components/Tabs/Tabs.spec.jsx
jondlm: Yeah, it's a bit of a crazy class name but it's descriptive heh

PR #65, opened 2016-03-03
Feature/grids

alisonailea:


src/components/Grid/Grid.jsx
```diff
@@ -0,0 +1,125 @@
...
+			children,
+			...passThroughs
+		} = this.props;
+
+		let scopedClasses = boundGridClassNames('~', propsToClass(this.props));
+
+		return (
+			<section className={classNames(className, scopedClasses)}
+				style={style}
+				{...passThroughs} >
```
jondlm: `passThroughs` should be spread at the head instead of the tail of `section`

src/components/Grid/Grid.jsx
```diff
@@ -0,0 +1,125 @@
...
+});
+
+/**
+ * {"categories": ["layout", "grid"], "extend": "Grid"}
+ *
+ * This component is designed to be used in Composits as a layout tool.
+ * You can use the Grid components themselves or create your own components using the Grid styles from Grid.less.
+ * Please see examples/CSS-Grid.jsx for more information.
+ */
+const Cell  = React.createClass({
```
jondlm: I think `Cell` should be a child of grid to conform to our existing patterns. @ogupte does that make sense here? Sorry I didn't spot this earlier, it shouldn't be a big refactor: 

```
<Grid>
  <Grid.Cell></Grid.Cell>
</Grid>
```

src/components/Grid/Grid.jsx
```diff
@@ -0,0 +1,125 @@
...
+				style={style}
+				{...passThroughs} >
+				{children}
+			</section>
+		);
+	}
+});
+
+/**
+ * {"categories": ["layout", "grid"], "extend": "Grid"}
```
jondlm: I don't think (not 100% sure) that docgen supports multiple components in a single file like this. We actually have another pattern by which we specify child components.

src/components/Grid/Grid.less
```diff
@@ -0,0 +1,66 @@
...
+	flex-shrink: 1;
+	padding: 10 0 0 (10/2);
+
+	&-is-full{
+		flex-grow: 0;
+		flex-shrink: 0;
+		flex-basis: 100%;
+	}
+	&-is-half{
+		flex-grow: 0;
```
jondlm: I read recently that the recommended way of writing flex is using the shorthand. So this would be `flex: 0 0 50%`. Not sure if it actually matters here or not.

jondlm: PS I read that in the [CSS Tricks article](https://css-tricks.com/snippets/css/a-guide-to-flexbox/)

> It is recommended that you use this shorthand property rather than set the individual properties. The short hand sets the other values intelligently.

src/components/Grid/Grid.jsx
alisonailea: It doesn't render it but `npm dev` fails if the second comment block isn't there.

src/components/Grid/Grid.less
alisonailea: your right. I had forgotten that article. thanks!

src/components/Grid/examples/JSX-Grid.jsx
```diff
@@ -0,0 +1,138 @@
...
+					</Cell>
+				</Grid>
+				<Grid is-gutterless>
+					<Cell is-half>
+						<p style={assign(gridStyle.half)}>gutterless half</p>
+				</Cell>
+				<Cell style={assign(gridStyle.quarter)} is-quarter>gutterless quarter</Cell>
+					{fillCells(1)}
+				</Grid>
+			</Grid>
```
ogupte: might help if we can break this example up in to a few smaller examples, so the source code blob is smaller. we can also be more explicit and not have to use helpers like `fillCells` to save space.

src/components/Grid/Grid.jsx
```diff
@@ -0,0 +1,124 @@
...
+		}, []);
+
+	if(_.isEmpty(classNameProps)){
+		return null
+	}
+
+	return classNameProps;
+}
+
+const Grid = React.createClass({
```
ogupte: we should be using the `createLucidComponentDefinition` within `React.createClass` which will also let u declare `Cell` a child component of `Grid`

danhahn:
Do we have any plans to be able to set a fixed width of a column.  So a user could set a with to be `250px`.  I could see a situation where that might be helpful.

alisonailea:
Dan, I just made several updates to the grid component, including more explicitly establishing the grid. Currently we are using a static 12 column grid, although it is worth adding to the backlog to make it flexible enough to work with a 24 col grid as well and the grid is flexbox flexible so it will setup a 12col grid inside any component of any width you want. So you could easily create a grid in a 250px column.

jondlm:
Hey @alisonailea make sure to fill out the `spec` file with at least the `common` tests.

- [ ] add test

PR #76, opened 2016-03-05
New `props` callback convention

jondlm:
- Updated CONTRIBUTING.md with new convention details
- Updated all controls with new convention
- Fixed tests
- Added new generic test for the new convention
- Added `{ event, props }` last arg to every control


src/components/Button/Button.jsx
```diff
@@ -81,14 +90,20 @@ const Button = React.createClass({
...
+		} = this.props;
+
+		if (!isDisabled) {
+			onClick({ uniqueId, event });
 		}
 	},
 
 	render() {
-		let {
+		const {
```
ogupte: this is a good change

src/components/ButtonGroup/ButtonGroup.jsx
```diff
@@ -107,7 +108,8 @@ const ButtonGroup = React.createClass(createLucidComponentDefinition({
 							isActive={_.includes(selectedIndices, index)}
 							{...buttonChildProp}
 							key={index}
-							onClick={_.partial(this.handleSelect, index, buttonChildProp.onClick)}
+							uniqueId={index}
```
ogupte: It might be confusing if you pass in `<ButtonGroup.Button uniqueId='foo'>` and when they click it, it comes back with an index. Maybe we should default to index, but use a passed in `uniqueId` if they specify.

jondlm: Yes, good catch. I totally agree.

jondlm: - [ ] fix ButtonGroup to handle children uniqueIds better
- [ ] fix RadioGroup to handle children uniqueIds better

src/components/TextField/TextField.jsx
```diff
@@ -84,8 +96,13 @@ const TextField = React.createClass({
...
 
 	handleChange(event) {
+		const {
+			uniqueId,
+			onChange,
+		} = this.props;
 		const value = _.get(event, 'target.value', '');
-		this.props.onChange(value, event);
+
+		onChange(value, { uniqueId, event });
```
ogupte: yep i think `uniqueId` was a good name for this

Issue #81, opened 2016-03-08
`Grid` not working correctly in IE11 and Safari


jondlm:
## Safari

![image](https://cloud.githubusercontent.com/assets/3290587/13617927/1309e50a-e536-11e5-9ea2-ab1f5467a741.png)

## IE 11

![image](https://cloud.githubusercontent.com/assets/3290587/13617937/2792fd72-e536-11e5-8dad-9cece64131e1.png)

alisonailea:
fixed in https://github.com/appnexus/lucid/pull/84

PR #84, opened 2016-03-09
#81 + #4

alisonailea:
Grids: Normalizing Flexbox for cross-browsers
These fixes are only needed until IE12 becomes standard. I also introduced a Normalize mixin that can be used in other components.
Also adding the checklist since it wasn't included in the original Grids PR
- [x] Manually tested across supported browsers
 - [x]  Chrome
 - [x]  Firefox
 - [x]  Safari
 - [x]  IE11
- [x]  Unit tests written (common at minimum)
- [x]  PR has one of the semver- labels
- [ ]  Two core team engineer approvals
- [ ]  One core team UX approval


src/styles/mixins.less
```diff
@@ -90,3 +90,457 @@
 	-webkit-transform:translate3d(0,0,0);
 }
 
+
```
ogupte: this code should exist as a separate source file and be included

src/docs/index.less
```diff
@@ -4,6 +4,11 @@
 
 @import 'http://yui.yahooapis.com/pure/0.6.0/pure-min.css';
 @import '../styles/variables';
+@import '../styles/mixins';
+
+// Normalize
+#Normalize > .html;
```
jondlm: Are these required for the grid to behave correctly?

src/styles/mixins.less
```diff
@@ -90,3 +90,457 @@
...
+/*! normalize.css v3.0.3 | MIT License | github.com/necolas/normalize.css */
+#Normalize {
```
jondlm: I'd suggest we keep this lowercase `#normalize` so it doesn't make people think Normalize is a react component (our existing convention).

src/styles/mixins.less
jondlm: Agreed, perhaps `mixins-normalize.less`

src/styles/mixins.less
nathaniellee: +1

src/docs/index.less
alisonailea: No but I wanted to normalize our demo page

src/styles/mixins.less
alisonailea: That makes sense. But I do want to differentiate a Mixin-group from a basic mixin

src/styles/mixins/normalize.less
```diff
@@ -0,0 +1,453 @@
+/*! normalize.css v3.0.3 | MIT License | github.com/necolas/normalize.css */
+#mixin-Normalize {
```
jondlm: Mixin is already implied by the use of a less namespace, also, could you make this non capitalized? It conflicts with an existing convention.

src/styles/mixins/normalize.less
```diff
@@ -0,0 +1,453 @@
+/*! normalize.css v3.0.3 | MIT License | github.com/necolas/normalize.css */
+#normalize {
```
alisonailea: switched name as per @jondlm suggestion

.nvmrc
```diff
@@ -0,0 +1 @@
+v4.2
```
alisonailea: https://github.com/creationix/nvm

Issue #78, opened 2016-03-08
Add progression variant to `Tabs`


jondlm:
Ref ANXR-312.

![progressivetabs_styling_update](https://cloud.githubusercontent.com/assets/3290587/13608041/d1eb753a-e507-11e5-9314-2487960768ad.jpg)

PR #85, opened 2016-03-10
Add progression variant to `Tabs` #78

danhahn:
## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [x] IE11


src/components/Tabs/Tabs.less
```diff
@@ -31,30 +31,74 @@
...
 
 		&:hover {
-			background-color: @color-gray-5;
+			background-color: @color-lightGray;
+			.Tabs-mixin-tab-arrow-background(@color-lightGray);
 		}
 
 		&:active {
-			box-shadow: @shadow-inset;
+			background-image: linear-gradient(#dbdbdb 0, @color-lightGray @Tabs-spacer-height);
```
jondlm: Looks like another shade of grey, variable?

src/components/Tabs/Tabs.less
```diff
@@ -31,30 +31,74 @@
...
 		}
 	}
-	&-Tab-is-active {
-		border-top: @Tabs-spacer-height solid @color-primary;
-		padding-top: 0;
-		color: #000;
-		cursor: initial;
 
+	&-Tab-is-disabled {
+		color: fade(#333, 50%);
```
jondlm: Seems like disabled text should be a global var

src/components/Tabs/Tabs.less
```diff
@@ -62,3 +106,19 @@
 		padding-top: 10px; // TODO: use our default gutter
 	}
 }
+
+.Tabs-mixin-tab-line(@color-Tab-line: @color-primary) {
```
jondlm: I don't think argument variables need to be scoped, this could just be `@color`

src/components/Tabs/Tabs.less
```diff
@@ -62,3 +106,19 @@
...
+	.lucid-Tabs-Tab-arrow {
+		.lucid-Tabs-Tab-arrow-tab-line {
+			fill: @color-Tab-line;
+		}
+	}
+}
+
+.Tabs-mixin-tab-arrow-background(@color-Tab-background: @color-white) {
```
jondlm: ditto

src/components/Tabs/examples/stateful-progression.jsx
```diff
@@ -0,0 +1,21 @@
...
+export default React.createClass({
+	render() {
+		return (
+			<div>
+				<StatefulTabs isOpen={false} isProgressive={true}>
+					<StatefulTabs.Tab Title='One'>One content</StatefulTabs.Tab>
+					<StatefulTabs.Tab Title='Two'>Two content</StatefulTabs.Tab>
+					<StatefulTabs.Tab Title='Three'>Three content</StatefulTabs.Tab>
+					<StatefulTabs.Tab Title='Disabled' isDisabled={true}>Disabled Content</StatefulTabs.Tab>
+					<StatefulTabs.Tab Title='Five'>Four content</StatefulTabs.Tab>
```
jondlm: Make this one disabled as well so the example is clear

src/components/Tabs/Tabs.less
danhahn: this was a little bit of hack.  when I change the `box-shodow` firefox did not like itl  So changed to a background gradient.  I pulled the colors off the JPG.  Since they were on the gray backgorund they were a little different than our standard grays.

src/components/Tabs/Tabs.less
danhahn: good point.  will change when I clean up some of the other variables issues that we have

src/components/Tabs/Tabs.less
danhahn: Just trying to keep it easy to read.  if they both are `@color` it might be a little confusing.

src/components/Tabs/Tabs.jsx
```diff
@@ -27,7 +27,13 @@ const Tabs = React.createClass(createLucidComponentDefinition({
 
 	childProps: {
 		Tab: {
-			isSelected: bool
+			isSelected: bool,
```
alisonailea: description for this prop?

src/components/Tabs/Tabs.jsx
```diff
@@ -103,12 +118,21 @@ const Tabs = React.createClass(createLucidComponentDefinition({
...
+									'Tab-is-active-and-open': isOpen && index === actualSelectedIndex,
+									'Tab-is-progressive': isProgressive && index !== tabChildProps.length - 1,
 								})}
 								key={index}
-								onClick={_.partial(onSelect, index)}
+								onClick={_.partial(onSelect, index, tabChildProp)}
 							>
 								{_.get(_.first(Tabs.Title.findInAllAsProps(tabChildProp)), 'children', '')}
+								{isProgressive && index !== tabChildProps.length - 1 ?
+										<svg className={boundClassNames('Tab-arrow')} xmlns='http://www.w3.org/2000/svg' width='8px' height='28px' viewBox='0 0 8 28' >
```
alisonailea: Should this be broken out into it's own SVG component?

src/components/Tabs/Tabs.less
```diff
@@ -12,53 +14,112 @@
...
 
 		&:hover {
-			background-color: @color-gray-5;
+			background-color: @color-lightGray;
+			.Tabs-mixin-tab-arrow-background(@color-lightGray);
 		}
 
 		&:active {
-			box-shadow: @shadow-inset;
+			background-image: linear-gradient(@color-gray 0, @color-lightGray @Tabs-spacer-height);
```
alisonailea: just a note to switch this to the new gradient mixin when Vertical List is merged

src/components/Tabs/Tabs.jsx
danhahn: Had it that way originally but because of an issue with the docs we can not have more than one component in a file.  Since it would only be used in the design it does not make sense to have a separate file.

Issue #86, opened 2016-03-10
Labeled radio button component


nathaniellee:
Still discussing whether this should be a new `LabeledRadioButton` composite component that leverages `RadioButton` or some other approach (I think we're all in agreement that we don't want to add `label` as a prop to `RadioButton`).

jondlm:
Would we want to include labeled versions of all our controls? This is a trickier question. The approach I took with `TextField` was to put an `Error` prop directly on it which leverages the `Validation` component under the hood. Should I have created a `ValidatedTextField` component instead?

nathaniellee:
I actually do think that it would have been better to go with a `ValidatableTextField` component instead.

jondlm:
Agreed! Will be doing that.

PR #92, opened 2016-03-10
#86 -- Added LabeledRadioButton component.

nathaniellee:
## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [x] IE11 (Win 7)
  - [x] Edge (Win 10)


src/components/LabeledRadioButton/LabeledRadioButton.jsx
```diff
@@ -0,0 +1,109 @@
...
+const {
+	bool,
+	func,
+	node,
+	object,
+	string,
+} = React.PropTypes;
+
+/**
+ * {"categories": ["controls", "toggles"]}
```
jondlm: Add `extend: "RadioButton"` as a property to this JSON and it will merge all RadioButton props with this.

src/components/LabeledRadioButton/LabeledRadioButton.jsx
```diff
@@ -0,0 +1,109 @@
...
+		 * Indicates that the component is in the "selected" state when true
+		 * and in the "unselected" state when false.
+		 */
+		isSelected: bool,
+
+		/**
+		 * Used to identify the purpose of this radio button to the user -- can
+		 * be any renderable content.
+		 */
+		label: node,
```
jondlm: Seems like this should use our child convention:

```
...
childProps: {
  Label: null,
}
...
```

jondlm:
The implementation seems a little odd to me. See how I did [ValidatedTextField](https://github.com/appnexus/lucid/blob/quality/90-fix-textfield/src/components/ValidatedTextField/ValidatedTextField.jsx) for a better example of what I'm thinking.

src/components/LabeledRadioButton/LabeledRadioButton.jsx
nathaniellee: There isn't a `Label` component at the moment... just a plain old regular prop. :)

src/components/LabeledRadioButton/LabeledRadioButton.less
```diff
@@ -0,0 +1,41 @@
...
+		cursor: not-allowed;
+	}
+
+	&:hover .lucid-RadioButton-visualization-container {
+		border-color: @color-primary;
+		transition: @RadioButton-animation-timing;
+	}
+
+	&:hover .lucid-RadioButton-visualization-glow {
+		opacity: 0.3;
```
ogupte: should be using the `.opacity(0.3)` mixin

src/components/LabeledRadioButton/LabeledRadioButton.less
```diff
@@ -0,0 +1,41 @@
...
+	}
+
+	&&-is-disabled .lucid-RadioButton-visualization-container,
+	&&-is-disabled:hover .lucid-RadioButton-visualization-container {
+		background-color: rgba(0, 0, 0, 0.05);
+		border-color: rgba(0, 0, 0, 0.1);
+	}
+
+	&&-is-disabled:hover .lucid-RadioButton-visualization-glow {
+		opacity: 0;
```
ogupte: should be using the `.opacity(0)` mixin

src/components/LabeledRadioButton/LabeledRadioButton.spec.jsx
```diff
@@ -0,0 +1,81 @@
...
+				const wrapper = shallow(
+					<LabeledRadioButton className='wut' isDisabled={true} isSelected={true} style={{ fontWeight: 'bold' }} onSelect={_.noop}
+							foo={1} bar={2} baz={3} qux={4} quux={5} />
+				);
+				const radioButtonProps = _.keys(wrapper.find(RadioButton).props());
+
+				// It should pass `foo`, `bar`, `baz`, `qux`, and `quux` through
+				// to the `RadioButton` instance.
+				_.forEach(['foo', 'bar', 'baz', 'qux', 'quux'], (prop) => {
+					assert(_.includes(radioButtonProps, prop));
```
ogupte: instead of using `_.includes` on `_.keys` just use `_.has` to check if the `prop` is on the object

src/components/LabeledRadioButton/examples/interactive.jsx
```diff
@@ -0,0 +1,64 @@
...
+
+export default React.createClass({
+	getInitialState() {
+		return {
+			flavor: 'vanilla'
+		};
+	},
+
+	handleSelectedChocolate() {
+		this.setState(_.assign({}, this.state, {
```
ogupte: the `_.assign` is redundant since `this.setState` already does an implicit assign

src/components/LabeledRadioButton/examples/states.jsx
```diff
@@ -0,0 +1,27 @@
...
+	marginRight: '5px'
+};
+
+export default React.createClass({
+	render() {
+		return (
+			<section style={{ display: 'flex' }}>
+				<LabeledRadioButton style={style} />
+				<LabeledRadioButton Label='String passed in as the label' style={style} />
+				<LabeledRadioButton Label={<p>HTML element passed in as the label</p>} style={style} />
```
ogupte: This example should be replaced with one that passes `Label` as a child element with children rather than passing an element thru props.

src/components/LabeledRadioButton/LabeledRadioButton.jsx
```diff
@@ -0,0 +1,116 @@
...
+				<RadioButton
+						className={className}
+						isDisabled={isDisabled}
+						isSelected={isSelected}
+						onSelect={onSelect}
+						{...passThroughs}
+				/>
+				{
+					labelChildProps
+							? labelChildProps.children || labelChildProps
```
ogupte: only `labelChildProps.children` is needed here since this logic is already handled (see https://github.com/appnexus/lucid/blob/master/src/util/child-component.js#L73)

src/components/LabeledRadioButton/LabeledRadioButton.spec.jsx
nathaniellee: Good call. Over engineering at its best... :)

src/components/LabeledRadioButton/LabeledRadioButton.less
nathaniellee: Didn't realize there was one. Nice. I don't think _any_ of our components are using that mixin right now...

src/components/LabeledRadioButton/LabeledRadioButton.jsx
jondlm: Oh cool, I didn't know that either!

src/components/LabeledRadioButton/LabeledRadioButton.jsx
```diff
@@ -0,0 +1,116 @@
...
+			onSelect,
+			style,
+			...passThroughs
+		} = this.props;
+
+		const labelChildProps = _.first(LabeledRadioButton.Label.findInAllAsProps(this.props));
+
+		return (
+			<label
+					className={classNames(boundClassNames('&', {
```
jondlm: Since @ogupte's refactor this can now become:

```
className={boundClassNames('&', { ... }, className)}
```

src/components/LabeledRadioButton/examples/states.jsx
nathaniellee: I think we should probably keep that _and_ add an example of children, no?

src/components/LabeledRadioButton/LabeledRadioButton.jsx
nathaniellee: As in not having to call `classNames`?

Issue #104, opened 2016-03-14
Labeled check box component


nathaniellee:
https://github.com/appnexus/lucid/issues/86... but with a check box. :)

PR #105, opened 2016-03-14
#104 -- Added LabeledCheckbox composite.

nathaniellee:
## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [x] IE11 (Win 7)
  - [x] Edge (Win 10)

### DIFF
--- a/src/components/LabeledCheckbox/LabeledCheckbox.less
+++ b/src/components/LabeledCheckbox/LabeledCheckbox.less
@@ -1,0 +1,41 @@
+.lucid-LabeledCheckbox {
+	align-items: center;
+	color: @color-darkGray;
+	cursor: pointer;
+	display: flex;
+	font: @font;
+	justify-content: flex-start;
+
+	&&-is-disabled {
+		color: @color-mediumGray;
+		cursor: not-allowed;
+	}
+
+	&:hover .lucid-Checkbox-visualization-container {
+		border-color: @color-primary;
+		transition: @Checkbox-animation-timing;
+	}
+
+	&:hover .lucid-Checkbox-visualization-glow {
+		.opacity(0.3);
+		transform: @Checkbox-glow-scale;
+	}
+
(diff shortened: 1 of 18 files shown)


src/components/LabeledCheckbox/LabeledCheckbox.less
```diff
@@ -0,0 +1,41 @@
...
+	justify-content: flex-start;
+
+	&&-is-disabled {
+		color: @color-mediumGray;
+		cursor: not-allowed;
+	}
+
+	&:hover .lucid-Checkbox-visualization-container {
+		border-color: @color-primary;
+		transition: @Checkbox-animation-timing;
```
jondlm: Might have some issues pulling in this variable since we are relying on the order of loading in this case.

jondlm:
Other than the less variable issue being a potential issue, LGTM!

jondlm:
Could you add the clarifications (IE 11, Edge) to the PULL_REQUEST_TEMPLATE?

src/components/LabeledCheckbox/LabeledCheckbox.less
nathaniellee: Is there any special sort of scoping introduced by the compilers? Less by itself would put `@Checkbox-animation-timing` into the global scope in which case it would be usable regardless of order.

nathaniellee:
I need to make a few adjustments to the styles as per a request from Ashley and David when UX reviewing the examples.

src/components/LabeledCheckbox/LabeledCheckbox.less
jondlm: It would put it into the global scope, but only if you load `Checkbox.less` before loading this file, otherwise it'll blow up.

nathaniellee:
I included the same styling tweaks to the `LabeledRadioButton` component in this branch. I felt that it didn't warrant a separate issue and pull request. Just an FYI to the rest of the team.

PR #111, opened 2016-03-16
Fix state issues with TextField, ValidatedTextField, and Tabs

### LINKED ISSUE #90: Fix up TextField
- [x] refactor out validation by creating another `ValidatedTextField` component
- [x] add reducers
- [x] add stateful example
- [x] ensure props can take over state when using the stateful version
- [x] add `onSubmit` callback with standard `(value, { event, props }) => {}` signature

jondlm:
## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [x] IE11 (Win 7)
  - [x] Edge (Win 10)
- [ ] Add one more test for `getProps` I think

There is a special case, the TextField, that needs to retain internal
state until new props are received. This change allows for that case
when its needed.

Also fixed the Tabs by not specifying an `isSelected` on one of the
children Tab items.


src/util/state-management.js
```diff
@@ -69,7 +81,8 @@ export function getStatefulPropsContext(reducers, { getState, setState }) {
 export function buildStatefulComponent(baseComponent, opts) {
 	opts = _.merge({
 		setStateWithNewProps: true, // if true, new props will update state, else prop has priority over existing state
-		replaceEvents: false // if true, function props replace the existing reducers, else they are invoked *after* state reducer returns
+		replaceEvents: false, // if true, function props replace the existing reducers, else they are invoked *after* state reducer returns
+		stateBeatsProps: false, // if true, when the component re-renders, state will take precedence over props, but on componentwillreceiveprops it will still favor props, there is a very limited case where this useful, mostly with text inputs to keep the user input until new props come in
```
jondlm: This is the new option. It's got a useful comment:

> when the component re-renders, state will take precedence over props, but on componentwillreceiveprops it will still favor props, there is a very limited case where this useful, mostly with text inputs to keep the user input until new props come in

jondlm:
Closing this PR since we're going to move state into the `TextField` instead of using the reducer pattern.

PR #117, opened 2016-03-17
#113 -- Added RadioGroup.Label feature.

nathaniellee:
## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [x] IE11 (Win 7)
  - [x] Edge (Win 10)


nathaniellee:
David reviewed this and approved via Slack.

src/components/RadioGroup/examples/default.jsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import RadioGroup from '../RadioGroup';
+
+const style = {
+	marginRight: '13px'
```
jondlm: Is this something we should be handling in our LESS files?

ogupte: i'm not sure? should we have styles in our docs less which are for specific examples?

jondlm: My question was more, should this spacing be default for labeled components.

nathaniellee: That's a good question. These are kind of arbitrary styles. The thing of it is, though, that it's up to the consumer to decide whether to render these in a horizontal or vertical fashion... would choosing default margins put more onus on the consumer to have to modify those styles?

I can understand if it's just a tiny bit of margin all around so they're not crushed up against each other (like how different browsers have their defaults around certain controls), but I do know that for the purposes of how they show up in our application David and Ashley expressed their idea of particular spacing sizes.

PR #126, opened 2016-03-18
Modal

jondlm:
## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [x] IE11 (Win 7) -- minor flex bug I'm working through
  - [x] Edge (Win 10)


risingspiral:
Is there a plan for an [x] icon as a way to cancel a modal or is that interaction pattern being eliminated?

Is there any notion of a "Are you sure you want to close this?" verification in this component?

jondlm:
@risingspiral the `x` pattern has been done away with per UX. I'm not sure about the verification step, let me drag @halseyhemingway in for that question.

src/util/generic-tests.jsx
```diff
@@ -7,7 +7,7 @@ import _ from 'lodash';
 import * as lucid from '../index';
 
 // Common tests for all our components
-export function common(Component, getDefaultProps=_.noop) {
+export function common(Component, { getDefaultProps = _.noop, selector = null } = {}) {
```
jondlm: I had to add a `selector` optional argument in case it's not possible to simply use the first element. This should help fix the common tests with `Banner` as well for #18 .

I moved `getDefaultProps` into an options object and gave it the same default value.

src/util/generic-tests.jsx
```diff
@@ -53,7 +57,7 @@ export function common(Component, getDefaultProps=_.noop) {
 				return acc.concat(node.prop('className').split(' '))
 			}, []);
 
-			let allClasses = parentClasses.concat(childrenClasses);
+			const allClasses = parentClasses.concat(childrenClasses);
```
jondlm: ¯\_(ツ)_/¯

src/styles/variables.less
```diff
@@ -3,6 +3,12 @@
 @fontFamily: "Helvetica Neue", Helvetica, Arial, Sans-Serif;
 @font: @fontSize @fontFamily;
 
+@size-font: 12px;
```
jondlm: I added these so we can get some consistency around font sizing. cc @danhahn

src/styles/mixins.less
```diff
@@ -113,3 +113,32 @@
...
+
+// -----------------------------------------------------------------------------
+// Transition group animations
+// -----------------------------------------------------------------------------
+
+// This set of mixins in intended to be used with ReactCSSTransitionGroup. It's
+// important that you place this mixin directly underneath the parent that
+// should be animating.
+
+.transition-group-animation-fade() {
```
jondlm: We might want to make the animation timing an argument

src/components/Modal/Modal.spec.jsx
```diff
@@ -0,0 +1,103 @@
...
+
+describeWithDOM('Modal', () => {
+	it('should fire the onEscape handler when escape is typed', () => {
+		const onEscape = sinon.spy();
+
+		mount(
+			<Modal isClosed={false} onEscape={onEscape} />
+		);
+
+		const event = new window.KeyboardEvent('keydown', {
```
jondlm: This actually works, I was so happy.

jondlm:
I've got most of the kinks worked out except for this really odd bug when you hover on a button in the `Modal.Footer`:

ie11-button-bug

nathaniellee:
I feel strongly that we should not conflate "modal" with "dialog". The term "modal" is an adjective describing a scenario in which the screen is blocked in such a way that the user is only able to interact with a very particular section of the interface. There's no such thing as "a modal" -- rather, UX is asking for a modal _dialog_ component.

We should have a `Modal` component that represents just the blocking overlay and then leverage that for a `Dialog` component that has an `isModal` prop.

halseyhemingway:
@risingspiral The "X" to close was indeed removed per UX's direction, specifically because we would then have to pop the "are you sure" type message every time since we would be unsure if they were closing to cancel or closing assuming it would save. 

There may be limited cases where we still want to ask if the user presses cancel (perhaps on screens like boolean group targeting where the user presumably did a lot of intricate work. While it's ultimately up to the application team to implement, they could probably utilize the small modal for that.

@nathaniellee would love to talk about the difference between Modal and Dialog. I think your proposal makes sense but want to make sure I understand the distinction. Thanks!

halseyhemingway:
modal_centered Hi @jondlm - one tiny visual nitpick before I throw the UX-approved label on. because of the scrollbar, when the modal is mathematically centered within the browser, it actually looks off-center. Can we account for the width of the scrollbar?

jondlm:
@halseyhemingway accounting for the scroll bar could cause more oddities. In some browsers (and depending if you're on mac/pc) the scroll bar will appear/disappear depending on how the content is scrolled. For example, most other libraries cheat and prevent the content behind the modal from scrolling by reaching outside the component and styling the root element of the page (`body`).

All that to say, I think we should leave the centering as is and not try to do anything hacky to prevent the content behind the modal from scrolling.

jondlm:
Beginning refactor to `Dialog` and `Modal`...

danhahn:
We should look at how we deal with the dialog when the screen size gets small.  If we have this over the top of a page and the window is too small the "action" button may be off the screen and there would be no way to dismiss the dialog.

jondlm:
@danhahn Here's what happens right now:

image

danhahn:
@jondlm I was getting some weirdness when using the mobile view in Chrome.  The same issue do not happen when I just resize my window.  There may be some issues when and if people view it on small screens. 

I would think at smaller sizes we might make it take up more of the screen on the left and right to get as much info on the screen as we can.

jondlm:
Just wrapped up my refactor into `Dialog` and `Overlay`. Overlay is the underlying component that powers Dialog. Basically it's possible now to have a `<Dialog isModal={false>` that won't block the user interaction with the rest of the screen. I did not implement draggable dialogs at this time.

src/components/Dialog/Dialog.jsx
```diff
@@ -0,0 +1,97 @@
...
+			<Overlay
+				{...passThroughs}
+				className={boundClassNames('&', className)}
+			>
+				<div
+					className={boundClassNames('&-window', {
+						'&-window-is-small': size === 'small',
+						'&-window-is-medium': size === 'medium',
+						'&-window-is-large': size === 'large',
+					})}
```
ogupte: you could do something like
```
boundClassNames.bind('&-window')('&', {
	'&-is-small': size === 'small',
	'&-is-medium': size === 'medium',
	'&-is-large': size === 'large',
})
```
if you want it to loose close to ur less src

ogupte: nvm ur less is more explicit. disregard

src/components/Dialog/Dialog.jsx
```diff
@@ -0,0 +1,97 @@
+import _ from 'lodash';
+import React from 'react';
+import Overlay from '../Overlay/Overlay';
+import { lucidClassNames } from '../../util/style-helpers';
+import { createLucidComponentDefinition }  from '../../util/component-definition';
+
+const boundClassNames = lucidClassNames.bind('&-Dialog');
+
```
nathaniellee: I'm wary of repeatedly used hard-coded values. Can we add
```
const LARGE = 'large';
const MEDIUM = 'medium';
const SMALL = 'small';
```
here and use those throughout this component?

src/components/Dialog/examples/1.small.jsx
```diff
@@ -0,0 +1,55 @@
...
+export default React.createClass({
+	getInitialState() {
+		return {
+			isShown: false
+		};
+	},
+
+	handleShow(isShown) {
+		this.setState({
+			isShown: isShown
```
nathaniellee:
```
this.setState({
  isShown
});
```
:)

src/components/Dialog/examples/1.small.jsx
```diff
@@ -0,0 +1,55 @@
...
+					isShown={this.state.isShown}
+					Header='Header'
+					size='small'
+				>
+					<div key={'info'}>
+						For better UX, we recommend NOT handling onEscape and
+						onOverlayClick when isModal is true. The term "modal" implies that
+						the user needs to interact with one of the buttons in the footer to
+						exit the dialog.
+					</div>
```
nathaniellee: Clever using the example to communicate development tips. :)

src/components/Overlay/Overlay.jsx
```diff
@@ -0,0 +1,149 @@
...
+			this.props.onOverlayClick({event, props: this.props });
+		}
+	},
+
+	render() {
+		const {
+			className,
+			isShown,
+			isModal,
+			portalId = 'Overlay-Portal-' + Math.random().toString(16).substr(2), // this was intentionally left out of getdefaultprops
```
nathaniellee: Can you explain why this isn't in `getDefaultProps`?

src/styles/variables.less
```diff
@@ -3,6 +3,15 @@
...
+@size-font-large: 18px;
+
+// Z Indices
+@zIndexModal: 4000;
+
+// Animations
+@timing-animation: 300ms;
```
nathaniellee: We might want to differentiate different timings as the duration of animation for the glow effect on `Checkbox`, `RadioButton` and `Switch` is lower.

src/util/generic-tests.jsx
nathaniellee: You _know_ I would have called you out on that! ;)

src/components/Overlay/Overlay.jsx
jondlm: I'll use the util. I was seeing some weird cases where it was generating the same portalId for several portals, I have no idea why it was doing that. I had it in default props but it wasn't working for some reason.

src/components/Dialog/Dialog.jsx
jondlm: Sure thing!

src/components/Dialog/examples/1.small.jsx
jondlm: Will do

src/styles/variables.less
jondlm: Yeah, perhaps I'll call this `@timing-animation-fade`

src/components/Overlay/Overlay.jsx
ogupte: that's because getDefaultProps runs only ones for all component instances :) i learned that the hard way too

src/components/Overlay/Overlay.jsx
```diff
@@ -0,0 +1,149 @@
...
+		 * Signature: `({ event, props }) => {}`
+		 */
+		onEscape: func,
+
+		/**
+		 * Fired when the user clicks on the background of the `Overlay`.
+		 *
+		 * Signature: `({ event, props }) => {}`
+		 */
+		onOverlayClick: func,
```
ogupte: this should probably be called `onBackgroundClick` since the _overlay_ includes the content which shouldn't trigger this when clicked.

src/components/Dialog/Dialog.jsx
jondlm: There's one place we shouldn't do this for documentation sake. Because our prop types are extracted from the AST, it can't read the value of the constant, so people would see `SMALL` in the docs. I'm going to leave it as strings in the propTypes but use the constants elsewhere.

src/components/Overlay/Overlay.jsx
jondlm: Thanks for the info, I'll stick it in state. That is good to know :+1:

Issue #109, opened 2016-03-15
Change button height for small button to 20px from 19px


danhahn:
Change the button height to be 20px not 19px like it is now.

jondlm:
Closed with #134

Issue #20, opened 2016-02-23
Minor refactor of unit tests for the Switch component


nathaniellee:
Tweaks in how they're built up following how I decided to do them for the in progress `RadioButton` component.

PR #134, opened 2016-03-22
#109 Change height from 19px to #20

danhahn:
## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [x] IE11 (Win 7)
  - [x] Edge (Win 10)

------

* Change height from 19px to 20px on small and short buttons
* Added new disabled selector to fix display issue after reset.css added

### DIFF
--- a/src/components/Button/Button.less
+++ b/src/components/Button/Button.less
@@ -18,11 +18,6 @@
 	line-height: @Button-size-lineHeight;
 	padding: 0 @Button-size-padding;
 
-	&[disabled] {
-		cursor: not-allowed;
-	}
-
-
 	// Types
 	&-primary {
 		color: @color-white;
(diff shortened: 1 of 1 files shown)


src/components/Button/Button.less
```diff
@@ -17,6 +18,10 @@
 	line-height: @Button-size-lineHeight;
 	padding: 0 @Button-size-padding;
 
+	&[disabled] {
```
jondlm: Shouldn't this be

```
&-is-disabled {}
```

Per [this class in the code](https://github.com/appnexus/lucid/blob/master/src/components/Button/Button.jsx#L111)

danhahn: yes and no.  There was an issue with 

```button[disabled], html input[disabled] {
    cursor: default;
}```

from `http://yui.yahooapis.com/pure/0.6.0/pure-min.css` which is our normalize css.

So I needed to add that so it would override that selector.

PR #136, opened 2016-03-22
Added Tag component.

nathaniellee:
## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [x] IE11 (Win 7)
  - [x] Edge (Win 10)


nathaniellee:
http://docspot.devnxs.net/projects/lucid/feature-ANXR-200-tag/#/components/Tag

nathaniellee:
I noticed that the "X" link is positioned a little too high in IE. :(

src/components/Tag/Tag.jsx
```diff
@@ -0,0 +1,96 @@
...
+					{..._.omit(passThroughs, 'onClose')}
+			>
+				{children}
+				{isCloseable
+						? (
+							<span
+									className={boundClassNames('&-close')}
+									onClick={this.handleClose}
+							>
+								{String.fromCharCode(0x00d7)}
```
ogupte: can you not just use the string `{'×'}` ?

src/components/Tag/Tag.spec.jsx
```diff
@@ -0,0 +1,97 @@
+import _ from 'lodash';
+import assert from 'assert';
+import React from 'react';
+import SyntheticEvent from 'react/lib/SyntheticEvent';
+import sinon from 'sinon';
+import { mount, shallow } from 'enzyme';
+shallow;
```
ogupte: this should be removed

src/components/Tag/Tag.spec.jsx
```diff
@@ -0,0 +1,97 @@
...
+
+		mount(<Tag isCloseable={true} onClose={onClose} />)
+				.find('.lucid-Tag-close')
+				.simulate(event);
+		assert(onClose.calledOnce);
+	}
+
+	describe('user clicks on the "X" button', () => {
+		it('calls the function passed in as the `onSelect` prop...', () => {
+			verifyOnSelect('click');
```
ogupte: logic should be inlined in the test, since it's only used once

src/components/Tag/Tag.spec.jsx
```diff
@@ -0,0 +1,97 @@
...
+		});
+
+		it('...and passes along a React synthetic event and the component props as part of the first argument.', () => {
+			verifyArguments('click');
```
ogupte: same here: logic should be inlined in the test, since it's only used once

src/components/Tag/Tag.jsx
```diff
@@ -0,0 +1,96 @@
...
+
+		/**
+		 * Passed through to the root element.
+		 */
+		style: object
+	},
+
+	getDefaultProps() {
+		return {
+			isCloseable: false,
```
ogupte: `isCloseable` should probably default to `true`

jondlm:
I'm not sure keyed into the discussion around this component, but should tags be nestable? It's fairly common in our apps to have "Applied Filters" with nested tags to display what filters are applied. I know there have been a couple attempts at tags/tag groups before and I want to make sure we get it right.

src/components/Tag/Tag.jsx
nathaniellee: Probably could... I just aped the `Banner` component since it seemed sensible for the two to share the same "X". :)

src/components/Tag/Tag.spec.jsx
nathaniellee: Oops. :\

src/components/Tag/Tag.spec.jsx
nathaniellee: For the time being I removed the test around touch events that I do in other components. I'm going to speak to Ashley and David about whether we should be consistent in covering those events with this component. If we do need to cover them, then I'll need to put those tests back in so I'm going to wait until after that conversation to decide whether I need to inline these functions. :)

src/components/Tag/Tag.jsx
nathaniellee: I'm iffy on that. If we look at it just from an internal usage perspective, then yeah it would make sense to make this true by default. However, I'd argue that the most common scenario "in the wild" where you'd see tags is on pages where you don't have the ability to remove them. In that sense I think the default would be not closeable...

nathaniellee:
I didn't work off of UX mocks explicitly for a `Tag` component. Rather, I worked off of mocks for tags used within a multi select component. I'm not sure whether David has come up with mocks for a general `Tag` component.

I pinged the team a day ago or so on the possibility of not exposing an explicit `Tag` component but rather offering a `TagGroup` component that could consist of one or more tags, driven by the nature of `ReactCSSTransitionGroup` -- would be good to chat about that tomorrow.

src/components/Tag/Tag.jsx
jondlm: There are pitfalls with react and putting characters directly into the source. I tried it with banner and it has issues some browser issues somewhere. Just leave the char code!

alisonailea: shouldn't we be using the close icon here?

src/components/Tag/examples/default.jsx
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+
+import Tag from '../Tag';
+
+export default React.createClass({
+	render() {
+		return (
+			<section>
+				<section>
```
alisonailea: There is nothing wrong with having a `<section>` inside a `<section>` but I expect a section to carry unique content. In this case I would replace the inner section tags with `<article>` instead.

src/components/Tag/Tag.less
```diff
@@ -0,0 +1,38 @@
+@Tag-spacing: 6px;
+
+.lucid-Tag {
+	background-color: @color-primary;
+	border-radius: 3px;
```
alisonailea: this should probably be `@border-radius-small` or something

src/components/Tag/Tag.less
```diff
@@ -0,0 +1,38 @@
...
+	font: @font;
+	line-height: 20px;
+	margin: 1px;
+	min-height: 20px;
+	padding: 0 @Tag-spacing;
+
+	&-close {
+		cursor: pointer;
+		fill: fade(@color-white, 50%);
+		font-size: 1rem;
```
alisonailea: I think we decided we are sticking to px or em

jondlm:
Going to close this one if/when the need arises again.

Issue #120, opened 2016-03-17
Panel Component


oliveiraddb:

jondlm:
Hey @oliveiraddb, could you give us a little more context as to what you're thinking? We like to get UX in front of us before we start building new components in Lucid so they fit into the larger framework in a meaningful way.

oliveiraddb:
Oh sorry. I'm suggesting a Panel like this one:
http://getbootstrap.com/components/#panels
I think it'd be good to have information organized in blocks when
necessary. It could have properties like title and children components like
Header, Body and Footer.

If you guys like the ideia but think it has low priority, let me know so
maybe i can create PR with that component.



2016-03-17 18:41 GMT-03:00 Jon de la Motte <notifications@github.com>:

jondlm:
No worries! Hey @halseyhemingway is there anything on the roadmap that a need for something like this?

halseyhemingway:
Yes it is! Its on JIRA as ANXR-258 (the UX team has been calling it a "container"). Its somewhere in the middle of the backlog at the current moment. Let me see if David can get us a quick mock. It should be pretty easy to knock out.

jondlm:
Awesome, sounds good. @oliveiraddb that'd be awesome if you want to tackle this once the mock up is ready. Make sure to read over the [contributing docs][cont] and let me know if you have any questions about implementation/coding style.

[cont]: https://github.com/appnexus/lucid/blob/master/CONTRIBUTING.md

oliveiraddb:
Great! I can do that!

If needed, I can handle issue #112 (Modal component) too as I see Modal as a Panel (container) with a close button and a layer behind. What do you guys think?

jondlm:
@oliveiraddb I'm actually almost done with `Modal`, there will be a PR for it later this afternoon :) And you're correct, they will look pretty similar likely.

PR #138, opened 2016-03-24
Panel Component #120

oliveiraddb:
## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [ ] Safari
  - [x] IE11 (Win 7)
  - [x] Edge (Win 10)


src/components/Container/Container.jsx
```diff
@@ -0,0 +1,88 @@
...
+import { createLucidComponentDefinition } from '../../util/component-definition';
+
+const boundClassNames = lucidClassNames.bind('&-Container');
+
+const {
+	node,
+	string
+} = React.PropTypes;
+
+function getHeader(headerChildProp, title){
```
jondlm: This is basically just a default value for the header, I'd scrap these functions

oliveiraddb: Sure, i'll handle it

src/components/Container/Container.jsx
```diff
@@ -0,0 +1,78 @@
...
+		/**
+		 * *Child Element* - Header contents. Only one `Header` is used.
+		 */
+		Header: node,
+		/**
+		 * *Child Element* - Header contents. Only one `Header` is used.
+		 */
+		Content: node,
+		/**
+		 * *Child Element* - Header contents. Only one `Header` is used.
```
nathaniellee: Probably want to edit the comments to refer to the props they're addressing. ;)

oliveiraddb: Oh, man! I haven't seen it before xD
Sure, I'll change it

src/components/Container/Container.jsx
```diff
@@ -0,0 +1,78 @@
...
+		const footerEl = (footerChildProp)?<footer {...footerChildProp} className={boundClassNames('&-footer')} />:null;
+
+		return (
+			<div className={boundClassNames('&', className)} style={style}>
+				{titleEl}
+				{headerEl}
+				{contentEl}
+				{footerEl}
+			</div>
+		)
```
nathaniellee: We generally don't abstract bits and pieces out of the `render`'s return expression since it introduces a lot of disconnect around what the markup structure ultimately looks like.
```
return (
	<div className={boundClassNames('&', className)} style={style}>
		{title ? (
			<header className={boundClassNames('&-header')}>
				<span className={boundClassNames('&-title')}>{title}</span>
			</header>
		) : null}
		{headerChildProp ? (
			<header {...headerChildProp} className={boundClassNames('&-header')} />
		) : null}
		{contentChildProp ? (
			<section {...contentChildProp} className={boundClassNames('&-content')} />
		) : null}
		{footerChildProp ? (
			<footer {...footerChildProp} className={boundClassNames('&-footer')} />
		) : null}
	</div>
);
```

src/components/Container/Container.less
```diff
@@ -0,0 +1,31 @@
+.lucid-Container {
+	overflow: hidden;
+	border-radius: 6px;
+	margin: 0 10px 10px 10px;
+	position: relative;
+	background-color: @color-white;
+	border: 1px solid @color-borderColor;
+
+	&-header {
+		padding: 12px;
```
nathaniellee: Prefer avoiding hard-coded values if they're repeatedly used throughout the file. I believe I've seen `12px` padding value in at least one other component before... might be worth discussing with UX whether this is a sort of standard that we can abstract into the `variables.less` file.

src/components/Container/Container.less
```diff
@@ -0,0 +1,31 @@
...
+		border-bottom: 1px solid @color-gray;
```
nathaniellee: If we know that the header and footer will always share the same bottom border we should probably abstract this out into a variable scoped to this component.

src/components/Container/Container.spec.jsx
```diff
@@ -0,0 +1,82 @@
...
+		});
+
+		it('should not exists', () => {
+			const wrapper = shallow(<Container><Container.Content>Content</Container.Content></Container>);
+			assert.equal(wrapper.find('.lucid-Container-header').length, 0);
+		});
+	});
+
+	describe('content', () => {
+		it('should exists', () => {
```
nathaniellee: Prefer these to be a bit more descriptive. When this and the next test run in the terminal I'm going to see

```
content
  should exists
  should not exists
```

That's confusing to me. :)

src/components/Container/Container.spec.jsx
```diff
@@ -0,0 +1,82 @@
...
+			const wrapper = shallow(<Container><Container.Content>Content</Container.Content><Container.Footer><button>Save</button></Container.Footer></Container>);
+			assert(wrapper.find('.lucid-Container-content'), `Content should include 'lucid-Container-content' class`);
+		});
+
+		it('should have the Container Footer class', () => {
+			const wrapper = shallow(<Container><Container.Content>Content</Container.Content><Container.Footer><button>Save</button></Container.Footer></Container>);
+			assert(wrapper.find('.lucid-Container-footer'), `Footer should include 'lucid-Container-footer' class`);
+		});
+	});
+});
```
nathaniellee: I'm not sure that we need unit tests around the presence of these CSS classes. May be worth a discussion.

nathaniellee:
I'm a bit confused... are the `title` prop and the `Header` child meant to represent essentially the same thing?

oliveiraddb:
We see both.. I imagine Header to be used as a toolbar, for example or another complex element, that's why i exposed it. But in case user just needs a simple title, the `title` prop could be used.

nathaniellee:
No love for Safari? ;)

src/components/Container/Container.spec.jsx
oliveiraddb: I based myself in another unit test i saw in project (Button, i think).. But maybe it's redundant here

src/components/Container/Container.spec.jsx
oliveiraddb: Got it!

oliveiraddb:
Yeah, sorry.. I have no Mac here so I couldn't test on it :/

nathaniellee:
A toolbar would be part of the content or if it needed to be rendered into the header would be part of `Header`'s children... and thus `Header` prop. :)

It seems redundant to have both of those...

nathaniellee:
```
<Container Header='Simple title' />
```
```
<Container>
  <Container.Header>Simple title</Container.Header>
</Container>
```
```
<Container>
  <Container.Header>
    <ul>
      <li>Please don't do this</li>
    </ul>
  </Container.Header>
</Container>
```
```
<Container Header={<ul><li>Please don't do this</li></ul>} />
```

oliveiraddb:
I see.. In Container case it's just a matter of styling. If you use Header, there's no other formatting besides padding but with `title` we have text formatting as specified on Jira. But in code, it's kind of redundant, i agree. Initially, it was specified just `title` but then Jon comment about using container in Modal, i decided to expose Header too. What do you suggest in this situation?

src/components/Container/Container.spec.jsx
alisonailea: I think we need to set up a way to test CSS but I don't think this is the place to do it.

nathaniellee:
IIRC there's going to be further discussion around what the `Container` component is actually supposed to be (the current mock shows it as "basically" a `Dialog`).

IMO the `Header` portion of the component should _not_ be styled as an 18px bold title like in the mockup but rather with our standard font styles. A `Dialog` component that is composed with the `Container` could then just pass whatever they want as part of the `Header` child/prop.

For example, a simple `Dialog`'s `render` could include something like...

```
<Container>
  <Container.Header>
    <h1>{this.props.title}</h1>
  </Container.Header>
</Container>
```

...where the `Dialog`'s styles could dictate what font styles are used for the `h1`.

src/components/Container/Container.jsx
```diff
@@ -0,0 +1,74 @@
...
+ * {"categories": ["layout"]}
+ *
+ * Container is used to wrap content to better organize elements in window.
+ */
+const Container = React.createClass(createLucidComponentDefinition({
+	displayName: 'Container',
+
+	childProps: {
+		Header: null,
+		Content: null,
```
jondlm: We don't need a `Content` child prop, it's simply anything that isn't a `Header` or `Footer`

src/components/Container/Container.jsx
```diff
@@ -0,0 +1,74 @@
...
+		const headerChildProp = _.first(Container.Header.findInAllAsProps(this.props));
+		const contentChildProp = _.first(Container.Content.findInAllAsProps(this.props));
+		const footerChildProp = _.first(Container.Footer.findInAllAsProps(this.props));
+
+		return (
+			<div className={boundClassNames('&', className)} style={style}>
+				{headerChildProp ? (
+					<header {...headerChildProp} className={boundClassNames('&-header')} />
+				) : null}
+				{contentChildProp ? (
```
jondlm: This can just be `{children}` when `Content` is removed

jondlm: Will probably want to wrap it in a `<div className={boundClassNames('&-content')` for styling reasons though

src/components/Container/Container.less
```diff
@@ -0,0 +1,27 @@
+@horizontal-lines: 1px solid @color-gray;
```
jondlm: These variable names aren't quite right. See the [contributing doc styling section](https://github.com/appnexus/lucid/blob/master/CONTRIBUTING.md#styling) for more examples of less variable names.

src/components/Container/Container.less
```diff
@@ -0,0 +1,27 @@
...
+@size-padding: 12px;
```
jondlm: `@Container-size-padding`

jondlm:
Sorry for the back and forth on this one but it's been argued that `Panel` is a better name :+1: So let's rename it back to `Panel`

src/components/Container/Container.jsx
```diff
@@ -0,0 +1,74 @@
...
+		 * Generally you should only have a single child element so the centering
+		 * works correctly.
+		 */
+		children: node
+	},
+
+	render: function() {
+		const {className, style, children} = this.props;
+
+		const headerChildProp = _.first(Container.Header.findInAllAsProps(this.props));
```
jondlm: Awesome, these are looking great!

src/components/Container/Container.spec.jsx
```diff
@@ -0,0 +1,46 @@
...
+import { common } from '../../util/generic-tests';
+
+import Container from './Container';
+
+describe('Container', () => {
+	common(Container);
+
+	describe('Header', () => {
+		it('renders a header element when included.', () => {
+			const wrapper = shallow(<Container><Container.Header>Header</Container.Header><Container.Content>Content</Container.Content></Container>);
```
jondlm: Could you wrap these lines better, here's [an example from dropmenu](https://github.com/appnexus/lucid/blob/master/src/components/DropMenu/DropMenu.spec.jsx#L41-L48)

jondlm:
You're super close on this one, it's shaping up nicely!

src/components/Container/Container.jsx
oliveiraddb: Quick question: what if User wants to insert an element outside content zonew

I.e.: &-content have a 12px padding but I want to insert a table that has to be full-width in container (no padding in parent element).

What have to be done in this situation?

jondlm: Good question: I think that's actually a pretty common case. I propose we add `isGutterless` to the top level props, add a `&-content-is-gutterless` conditional class, and remove the padding for that class.

PR #147, opened 2016-03-26
Feature/autocomplete

ogupte:
## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [x] IE11 (Win 7)
  - [x] Edge (Win 10)


halseyhemingway:
UX Approved as pertains to Project Periscope. Since the UX for this component originated from outside of the core Lucid team, I'd like to refine it a bit (or at least make sure I understand the design decisions) before widely promoting it.

src/components/Autocomplete/Autocomplete.less
```diff
@@ -0,0 +1,83 @@
...
+			fill: @color-black;
+			width: 8px;
+			height: 8px;
+			margin-left: auto;
+		}
+
+		// States
+
+		&-is-expanded {
+			border-color: @featured-color-primary-borderColor;
```
danhahn: Should just be `@color-primary`

src/components/Autocomplete/Autocomplete.less
```diff
@@ -0,0 +1,83 @@
...
+
+	&-Control {
+		.no-safari-flicker();
+		.box-sizing();
+		//.gradient-animation( @featured-color-default-gradientStartColor, @featured-color-default-gradientEndColor);
+		cursor: pointer;
+		height: 28px;
+		display: flex;
+		outline: none;
+		border: 1px solid @featured-color-default-borderColor;
```
danhahn: Should this be just be `@color-borderColor`?

ogupte: yeah i'll take out the commented code too

src/components/Autocomplete/Autocomplete.less
```diff
@@ -0,0 +1,83 @@
...
+		border-radius: @Button-size-borderRadius;
+		font: @font;
+		color: @color-black;
+		letter-spacing: .05em;
+		line-height: @Button-size-lineHeight;
+		padding: 0 10px;
+		align-items: center;
+
+		//.lucid-Autocomplete-Control-content {
```
danhahn: If we are not using this this get rid of it

src/components/Autocomplete/Autocomplete.less
```diff
@@ -0,0 +1,83 @@
+.lucid-Autocomplete {
+
+	&-Control {
+		.no-safari-flicker();
+		.box-sizing();
+		//.gradient-animation( @featured-color-default-gradientStartColor, @featured-color-default-gradientEndColor);
```
danhahn: If we are not using this get rid of it

src/components/Autocomplete/Autocomplete.jsx
```diff
@@ -0,0 +1,291 @@
...
+	bool,
+	func,
+	object,
+	shape,
+	string
+} = React.PropTypes;
+
+/**
+ *
+ * {"categories": ["controls", "text"], "madeFrom": ["DropMenu"]}
```
jondlm: Add `CaretIcon` to `madeFrom`

src/components/Autocomplete/Autocomplete.jsx
```diff
@@ -0,0 +1,291 @@
...
+				<DropMenu.Control onClick={this.handleControlClick}>
+					<div className={boundClassNames('&-Control', {
+						'&-Control-is-expanded': isExpanded,
+						'&-Control-is-disabled': isDisabled
+					})}>
+						<input
+							{..._.omit(passThroughs, ['onChange', 'onSelect', 'onExpand', 'value', 'children'])}
+							type='text'
+							className={boundClassNames('&-Control-input')}
+							ref='inputNode'
```
jondlm: I'd leave this, but FWIW I think strings for refs are going away in favor of using a function that attaches something to the class.

ogupte: i originally had a function, but i changed to a string so that it wouldn't be invoked more than it needed to be, since i just need to keep a reference to the DOMNode once anyways. I don't mind changing it back later.

src/components/Autocomplete/Autocomplete.jsx
```diff
@@ -0,0 +1,291 @@
...
+		}
+	},
+
+	getInputValue() {
+		return _.get(this.refs, 'inputNode.value', this.props.value);
+	},
+
+	setInputValue(value) {
+		if (this.refs.inputNode) {
+			this.refs.inputNode.value = value;
```
nathaniellee: From the React docs:
> If you want to preserve Google Closure Compiler advanced-mode crushing resilience, make sure to never access as a property what was specified as a string.

ogupte: probably premature at this point

src/components/Autocomplete/Autocomplete.jsx
nathaniellee: > Also note that when writing refs with inline function expressions as in the examples here, React sees a different function object each time so on every update, ref will be called with null immediately before it's called with the component instance.

Yeah... not sure we want to have that function called over and over again.

src/components/Autocomplete/examples/1.basic.jsx
```diff
@@ -0,0 +1,23 @@
...
+					'Portland',
+					'portal',
+					'porridge',
+					'potent',
+					'please'
+				]}
+			/>
+		);
+	}
+});
```
nathaniellee: Seems like the perfect opportunity to use a pure function component... ;)

```
export default () => (
  <Autocomplete
    placeholder='Enter a word...'
    suggestions={[
      'Portland',
      'portal',
      'porridge',
      'potent',
      'please'
    ]}
  />
);
```

src/components/Autocomplete/Autocomplete.jsx
nathaniellee: Really only if we're still undecided as to whether to stick with `ref='foo'` or go with `ref={function () { ... }}`...

src/components/Autocomplete/examples/1.basic.jsx
jondlm: That's not how we write our docs, consistency is better IMO.

src/styles/components.less
```diff
@@ -29,4 +29,4 @@
 @import '../components/Tabs/Tabs';
 @import '../components/Overlay/Overlay';
 @import '../components/Dialog/Dialog';
-
+@import '../components/Autocomplete/Autocomplete'; // Should import after DropMenu
```
jondlm: We should hoist the variables into `variable.less` rather than continuing to do this. Filed #148 for tech debt.

PR #159, opened 2016-03-31
Gh-pages support

### LINKED ISSUE #151: Setup gh-pages with the latest docs page
It's not ideal, since gh-pages doesn't support versioning, but we should at least provide the latest docs via `gh-pages`.

jondlm:
## PR Checklist


jondlm:
I fully expect this to have some kinks when I roll it out to master. Seems like these things always do!

PR #160, opened 2016-03-31
Added ChevronIcon and Expander components.

nathaniellee:
Docspot demo: http://docspot.devnxs.net/projects/lucid/feature-ANXR-190-expand-collapse/#/components/Expander

## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [x] IE11 (Win 7)
  - [x] Edge (Win 10)

### DIFF
--- a/src/components/Expander/Expander.jsx
+++ b/src/components/Expander/Expander.jsx
@@ -82,21 +82,33 @@ const Expander = React.createClass(createLucidComponentDefinition({
 		};
 	},
 
+	componentWillMount() {
+		this._labelKey = 0;
+	},
+
+	componentWillReceiveProps(nextProps) {
+		const currentLabel = _.first(Expander.Label.findInAllAsProps(this.props)).children;
+		const nextLabel = _.first(Expander.Label.findInAllAsProps(nextProps)).children;
+
+		if (currentLabel !== nextLabel) {
+			this._labelKey++;
+		}
+	},
+
 	render() {
 		const {
 			children,
 			className,
 			isExpanded,
-			onExpand,
 			style,
 			...passThroughs
 		} = this.props;
 
 		const labelChildProps = _.first(Expander.Label.findInAllAsProps(this.props));
-		const labelKey = isExpanded ? 'expandedLabel' : 'unexpandedLabel';
 
 		return (
 			<div
+				{..._.omit(passThroughs, 'onExpand')}
 				className={boundClassNames('&', {
 					'&-is-expanded': isExpanded,
 				}, className)}
@@ -117,7 +129,7 @@ const Expander = React.createClass(createLucidComponentDefinition({
 					>
 						{
 							labelChildProps
-									? <span key={labelKey}>{labelChildProps.children || labelChildProps}</span>
+									? <span key={this._labelKey}>{labelChildProps.children || labelChildProps}</span>
 									: null
 						}
 					</ReactCSSTransitionGroup>
--- a/src/components/Expander/Expander.jsx
+++ b/src/components/Expander/Expander.jsx
@@ -124,7 +124,6 @@ const Expander = React.createClass(createLucidComponentDefinition({
 						transitionName={boundClassNames('&-text')}
 						transitionEnterTimeout={100}
 						transitionLeaveTimeout={100}
-						style={{position: 'relative'}}
 						className={boundClassNames('&-text')}
 					>
 						{
--- a/src/components/Expander/Expander.jsx
+++ b/src/components/Expander/Expander.jsx
@@ -61,7 +61,7 @@ const Expander = React.createClass(createLucidComponentDefinition({
 		 *
 		 * Signature: `(isExpanded, { event, props }) => {}`
 		 */
-		onExpand: func,
+		onToggle: func,
 
 		/**
 		 * Passed through to the root element.
@@ -78,7 +78,7 @@ const Expander = React.createClass(createLucidComponentDefinition({
 	getDefaultProps() {
 		return {
 			isExpanded: false,
-			onExpand: _.noop
+			onToggle: _.noop
 		};
 	},
 
@@ -104,11 +104,11 @@ const Expander = React.createClass(createLucidComponentDefinition({
 			...passThroughs
 		} = this.props;
 
-		const labelChildProps = _.first(Expander.Label.findInAllAsProps(this.props));
+		const labelChildProp = _.first(Expander.Label.findInAllAsProps(this.props));
 
 		return (
 			<div
-				{..._.omit(passThroughs, 'onExpand')}
+				{...passThroughs}
 				className={boundClassNames('&', {
 					'&-is-expanded': isExpanded,
 				}, className)}
@@ -126,11 +126,9 @@ const Expander = React.createClass(createLucidComponentDefinition({
 						transitionLeaveTimeout={100}
 						className={boundClassNames('&-text')}
 					>
-						{
-							labelChildProps
-									? <span key={this._labelKey}>{labelChildProps.children || labelChildProps}</span>
-									: null
-						}
+						{labelChildProp ?
+							<span key={this._labelKey}>{labelChildProp.children}</span>
+						: null}
 					</ReactCSSTransitionGroup>
 				</header>
 				<section className={boundClassNames('&-content', {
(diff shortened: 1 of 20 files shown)


nathaniellee:
I'm having some issues with the width and height of the fake button that the `ChevronIcon` is in specifically on Edge. I'm wondering if there's some sort of bug around how that browser handles `box-sizing`.

src/components/Expander/Expander.jsx
```diff
@@ -0,0 +1,152 @@
...
+		/**
+		 * Passed through to the root element.
+		 */
+		style: object,
+
+		/**
+		 * Child element whose children represents content to be shown next to
+		 * the expander icon.
+		 */
+		Label: any
```
jondlm: We might not need this anymore since @ogupte has fixed the docs generator

ogupte: we do still need this for the time being, i need make the docs parsing better to support lucid child components

src/components/Expander/Expander.jsx
```diff
@@ -0,0 +1,152 @@
...
+			isExpanded,
+			style,
+			...passThroughs
+		} = this.props;
+
+		const labelChildProps = _.first(Expander.Label.findInAllAsProps(this.props));
+
+		return (
+			<div
+				{..._.omit(passThroughs, 'onExpand')}
```
ogupte: is this omit necessary? `div` probably ignores props it doesn't recognize.

src/components/Expander/Expander.jsx
```diff
@@ -0,0 +1,152 @@
...
+					</span>
+					<ReactCSSTransitionGroup
+						transitionName={boundClassNames('&-text')}
+						transitionEnterTimeout={100}
+						transitionLeaveTimeout={100}
+						className={boundClassNames('&-text')}
+					>
+						{
+							labelChildProps
+									? <span key={this._labelKey}>{labelChildProps.children || labelChildProps}</span>
```
ogupte: Only `labelChildProps.children` is needed here. `labelChildProps` is always a props object with a `children` prop.

src/components/Expander/Expander.jsx
```diff
@@ -0,0 +1,152 @@
...
+					'&-content-is-expanded': isExpanded
+				})}>
+					{children}
+				</section>
+			</div>
+		);
+	},
+
+	handleExpand(event) {
+		this.props.onExpand(!this.props.isExpanded, {
```
ogupte: onExpand shouldn't have to pass the `!this.props.isExpanded` payload. the reducer/handler should already have that data.

src/components/Expander/Expander.reducers.js
```diff
@@ -0,0 +1,6 @@
+export function onExpand(state = {}, isExpanded) {
+	return {
+		...state,
+		isExpanded
+	};
+};
```
ogupte: should be:
```
onExpand(state) {
  return {
    ...state,
    isExpanded: !state.isExpanded
  };
}
```

src/components/Expander/Expander.jsx
jondlm: I think we've been passing it as a convenience. 

[Other example](https://github.com/appnexus/lucid/blob/master/src/components/Switch/Switch.jsx#L114)
[Another example](https://github.com/appnexus/lucid/blob/master/src/components/Checkbox/Checkbox.jsx#L118)

src/components/Expander/Expander.reducers.js
ogupte: May want to rethink the name/behavior of this reducer/handler. The name `onExpand` implies that `isExpanded` will always be `true` when it returns, which is not the case, currently.

I suggest you either:

1. add an `onCollapse` handler/reducer which is called if `isExpanded` is true
2. change the name to `onToggle` or something more appropriate to the behavior

src/components/Expander/Expander.jsx
ogupte: We should probably get some consistency on this pattern. In DropMenu, `onExpand` and `onCollapse` send along no payload, see here:
https://github.com/appnexus/lucid/blob/master/src/components/DropMenu/DropMenu.reducers.js#L3

src/components/Expander/Expander.spec.jsx
```diff
@@ -0,0 +1,134 @@
...
+	});
+});
+
+describeWithDOM('Expander', () => {
+	describe('user clicks on the header', () => {
+		it('calls the function passed in as the `onExpand` prop...', () => {
+			_.forEach(['-header', '-icon', '-text'], (classSubstring) => {
+				const onExpand = sinon.spy();
+
+				mount(<Expander onExpand={onExpand} />).find(`.lucid-Expander${classSubstring}`).simulate('click');
```
ogupte: as a good practice you should call `.unmount()` on these wrappers which use the DOM. I might not be necessary in your case, but some components do cleanup tasks in the DOM during the `componentWillUnmount` lifecycle method, which won't be invoked if you don't call `.unmount()`.

For Example, anything which uses Portal depends on this this lifecycle method to cleanup the DOM for the next test. Again, might not be necessary for this component, but it's a good habit to call `.unmount()` in an `afterEach` section :)

src/components/Expander/Expander.spec.jsx
```diff
@@ -0,0 +1,134 @@
...
+				});
+			});
+		});
+	});
+});
+
+describeWithDOM('Expander', () => {
+	describe('user clicks on the header', () => {
+		it('calls the function passed in as the `onExpand` prop...', () => {
+			_.forEach(['-header', '-icon', '-text'], (classSubstring) => {
```
ogupte: You could test all these cases in just one `mount()`ed component, with only one spy, and it might be easier to read too.

src/components/Expander/Expander.spec.jsx
```diff
@@ -0,0 +1,134 @@
...
+			});
+		});
+
+		it('...and when `isExpanded` equals `true` passes along `false` as the first argument and a React synthetic event as the second argument.', () => {
+			_.forEach(['-header', '-icon', '-text'], (classSubstring) => {
+				const onExpand = sinon.spy();
+
+				mount(<Expander isExpanded={true} onExpand={onExpand} />).find(`.lucid-Expander${classSubstring}`).simulate('click');
+				assert.equal(onExpand.args[0][0], false);
+				assert(_.last(onExpand.args[0]).event instanceof SyntheticEvent);
```
ogupte: I think the `instanceof` check isn't necesarry, just the fact that event was passed should be enough to make this test pass. Some components don't even pass SyntheticEvents here which is OK too. (Autocomplete's `onChange` is [one example](https://github.com/appnexus/lucid/blob/master/src/components/Autocomplete/Autocomplete.jsx#L189))

ogupte:
`LabeledSwitch` changes should be in a different PR.

src/components/Expander/Expander.jsx
```diff
@@ -0,0 +1,152 @@
...
+	componentWillMount() {
+		this._labelKey = 0;
+	},
+
+	componentWillReceiveProps(nextProps) {
+		const currentLabel = _.first(Expander.Label.findInAllAsProps(this.props)).children;
+		const nextLabel = _.first(Expander.Label.findInAllAsProps(nextProps)).children;
+
+		if (currentLabel !== nextLabel) {
+			this._labelKey++;
```
ogupte: It's sad that this is necessary :(

src/components/Expander/Expander.reducers.js
jondlm: I agree that the name isn't the best. `onToggle` seems more appropriate.

src/components/Expander/Expander.jsx
jondlm: Agreed, we couldn't find any sane way to detect if the content changed :(

src/components/Expander/Expander.jsx
jondlm: I think we're okay so far. I think we _should_ provide the boolean in the case where the callback is a toggle (like in this case). In the _DropMenu_ example I agree that it doesn't make sense since you have a callback for each of the two states.

src/components/Expander/Expander.jsx
nathaniellee: Good call. LOL

src/components/Expander/Expander.spec.jsx
jondlm: We actually have a common test for this: https://github.com/appnexus/lucid/blob/master/src/util/generic-tests.jsx#L113

PR #162, opened 2016-04-01
Make Autocomplete input style selector more specific.

ogupte:
Addresses #161

### DIFF
--- a/src/components/Autocomplete/Autocomplete.less
+++ b/src/components/Autocomplete/Autocomplete.less
@@ -3,7 +3,7 @@
 
 .lucid-Autocomplete {
 
-	&-Control {
+	& &-Control {
 		.no-safari-flicker();
 		.box-sizing();
 		cursor: pointer;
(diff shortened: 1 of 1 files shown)


src/components/Autocomplete/Autocomplete.less
```diff
@@ -19,13 +19,15 @@
 		padding: 0 10px;
 		align-items: center;
 
-		&-input {
-			background: none;
-			border: none;
+		& .lucid-Autocomplete-Control-input {
+			background: none transparent;
```
jondlm: Use `@color-transparent`

src/components/Autocomplete/Autocomplete.less
```diff
@@ -3,7 +3,7 @@
 
 .lucid-Autocomplete {
 
-	&-Control {
+	& &-Control {
```
jondlm: Throw a comment in for why this has to be more specific

ogupte: Comment added near the input selector below.

PR #163, opened 2016-04-01
Add new util functions for child-component for element filtering

ogupte:


src/util/child-component.js
```diff
@@ -103,3 +89,43 @@ export function createChildComponent (definition={}) {
...
+	return _.filter(React.Children.toArray(children), (node) => (React.isValidElement(node) && _.includes(elementTypes, node.type)));
+}
+
+export function getChildComponentPropsArray(props, childComponentType) {
+	const propName = _.get(childComponentType, 'childComponentDefinition.propName');
+	if (propName) {
+		const currentProps = _.chain(props)
+			.get(propName) // grab the prop we care about, e.g. "Child"
+			.thru(x => [x]) // wrap in an array
+			.flatten()
```
nathaniellee: What's happening at this step...?

src/util/child-component.js
```diff
@@ -103,3 +89,43 @@ export function createChildComponent (definition={}) {
...
+		const childComponentPropsArray = _.map(currentProps, (childProp) => {
+			if (_.isPlainObject(childProp)) {
+				return childProp;
+			} else {
+				return {
+					children: childProp
+				};
+			}
+		});
+		return childComponentPropsArray;
```
nathaniellee: Boo to `if` blocks! :)

```
return _.map(currentProps, (childProp) => _.isPlainObject(childProp)
    ? childProp
    : {
      children: childProp
    };
```

src/util/child-component.js
```diff
@@ -103,3 +89,43 @@ export function createChildComponent (definition={}) {
...
+		const childComponentProps = getChildComponentPropsArray(props, childComponentType);
+		return acc.concat(_.map(childComponentProps, (childComponentProp) => ({
+			type: childComponentType,
+			props: childComponentProp
+		})));
+	}, []).concat(_.map(filterElementsByType(props.children, childComponentTypes), (element) => ({
+		type: element.type,
+		props: element.props
+	})));
+}
```
nathaniellee: Function passed into `_.map`:

```
({ props, type }) => ({ props, type })
```

src/util/child-component.spec.jsx
```diff
@@ -330,3 +333,187 @@ describe('#rejectNullElements', () => {
...
+					assert(_.has(itemProps, 'children'));
+				});
+
+				return (
+					<ul className='selection'>
+						{_.map(itemsProps, (itemProps, i) => (
+							<li className={'item' + (itemProps.isAvailable ? ' available' : '')} key={i}>
+								{itemProps.children}
+							</li>
+						))}
```
nathaniellee:
```
{_.map(itemsProps, ({ children, isAvailable }, i) => (
  <li className={'item' + (isAvailable ? ' available' : '')} key={i}>
    {children}
  </li>
))}
```

src/util/child-component.js
jondlm: Meh, ternaries are not always more readable

src/util/child-component.js
jondlm: I wrote that and I can't remember why. I remember it's critical though :/

jondlm: I think it's because you can have multiple children or a single prop and we want to flatten either of those into a regular array.

jondlm:
Cool, I think I've needed this elsewhere before. Thanks for putting it in.

src/util/child-component.js
nathaniellee: Gotcha. Makes sense. :)

src/util/child-component.js
nathaniellee: As long as the branched expressions aren't gigantic and you format them well, they're _beautiful_. ;)

src/util/child-component.js
ogupte: actually, now that i think about it, the map isn't really needed...i like the fact that it might be an actual element because you'll know if that component came from children since `React.isValidElement` would return true for it, so u kind of get extra data for free

ogupte:
yeah, i needed to pull out certain types of child components, but i wanted to maintain the rendering-order of those components, i can probably go back for some of our other components and use this now instead of how i was doing it before.

PR #169, opened 2016-04-05
Panel fixes

### LINKED ISSUE #166: Fix Panel example
Fix the _No Padding In Content_ example by swapping the table with some text. We don't want to put `pure-table`s in our examples since pure is only temporarily used for our demo page and isn't the support way of doing tables with our library.

### LINKED ISSUE #168: Panel Header and Footer should pass through className

jondlm:
Fixes #168 and #166

## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [x] IE11 (Win 7)
  - [x] Edge (Win 10)

### DIFF
--- a/src/components/Panel/Panel.jsx
+++ b/src/components/Panel/Panel.jsx
@@ -50,10 +50,10 @@ const Panel = React.createClass(createLucidComponentDefinition({
 
 	render: function() {
 		const {
-			className,
-			style,
 			children,
-			isGutterless
+			className,
+			isGutterless,
+			style,
 		} = this.props;
 
 		const headerChildProp = _.first(Panel.Header.findInAllAsProps(this.props));
@@ -69,7 +69,7 @@ const Panel = React.createClass(createLucidComponentDefinition({
 				{headerChildProp ? (
 					<header
 						{...headerChildProp}
-						className={boundClassNames('&-header', headerChildProp.className)}
+						className={boundClassNames('&-Header', headerChildProp.className)}
 					/>
 				) : null}
 
@@ -80,7 +80,7 @@ const Panel = React.createClass(createLucidComponentDefinition({
 				{footerChildProp ? (
 					<footer
 						{...footerChildProp}
-						className={boundClassNames('&-footer', footerChildProp.className)}
+						className={boundClassNames('&-Footer', footerChildProp.className)}
 					/>
 				) : null}
 			</div>
--- a/src/components/Panel/Panel.jsx
+++ b/src/components/Panel/Panel.jsx
@@ -49,22 +49,39 @@ const Panel = React.createClass(createLucidComponentDefinition({
 	},
 
 	render: function() {
-		const {className, style, children, isGutterless} = this.props;
+		const {
+			className,
+			style,
+			children,
+			isGutterless
+		} = this.props;
 
 		const headerChildProp = _.first(Panel.Header.findInAllAsProps(this.props));
 		const footerChildProp = _.first(Panel.Footer.findInAllAsProps(this.props));
-		const classNames = boundClassNames('&', className, (isGutterless)?'&-content-is-gutterless':null);
 
 		return (
-			<div className={classNames} style={style}>
+			<div
+				className={boundClassNames('&', className, {
+					'&-is-not-gutterless': !isGutterless
+				})}
+				style={style}
+			>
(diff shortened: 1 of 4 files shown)


src/components/Panel/Panel.spec.jsx
```diff
@@ -19,6 +19,16 @@ describe('Panel', () => {
 			assert.equal(wrapper.find('.lucid-Panel-header').length, 1);
 		});
 
+		it('passes through className', () => {
+			const wrapper = mount(
```
ogupte: All these tests should be using `shallow` instead of `mount` since none of them make any assertions on the DOM. Also as a rule, when we use `mount`, be sure to use the `describeWithDOM` helper instead of the regular `describe`.

src/components/Panel/Panel.spec.jsx
```diff
@@ -42,6 +52,16 @@ describe('Panel', () => {
...
+					<Panel.Footer className='bar'>Footer</Panel.Footer>
+					Content
+				</Panel>
+			);
+			assert.equal(wrapper.find('.lucid-Panel-footer.bar').length, 1);
+		});
+
 		it('does not render a footer element when not included.', () => {
 			const wrapper = mount(
 				<Panel>
```
ogupte: component Spec files should follow the pattern:
```
component {
  render {...}
  props {...}
  childComponents {...}
}
```

src/components/Panel/Panel.jsx
```diff
@@ -49,22 +49,39 @@ const Panel = React.createClass(createLucidComponentDefinition({
...
+
+				<section className={boundClassNames('&-content')} >
+					{children}
+				</section>
+
 				{footerChildProp ? (
-					<footer {...footerChildProp} className={boundClassNames('&-footer')} />
+					<footer
+						{...footerChildProp}
+						className={boundClassNames('&-footer', footerChildProp.className)}
```
ogupte: use capital 'F' in `&-Footer` since `Footer` is a child component

src/components/Panel/Panel.jsx
```diff
@@ -49,22 +49,39 @@ const Panel = React.createClass(createLucidComponentDefinition({
...
+				style={style}
+			>
 				{headerChildProp ? (
-					<header {...headerChildProp} className={boundClassNames('&-header')} />
-				) : null}
-				{children ? (
-					<section className={boundClassNames('&-content')}>{children}</section>
+					<header
+						{...headerChildProp}
+						className={boundClassNames('&-header', headerChildProp.className)}
```
ogupte: use capital 'H' in `&-Header` since `Header` is a child component

src/components/Panel/Panel.spec.jsx
jondlm: We should put that in the CONTRIBUTING docs

src/components/Panel/Panel.jsx
jondlm: Good call

src/components/Panel/Panel.spec.jsx
jondlm: Ahh I missed that

src/components/Panel/Panel.jsx
```diff
@@ -49,22 +49,39 @@ const Panel = React.createClass(createLucidComponentDefinition({
 	},
 
 	render: function() {
-		const {className, style, children, isGutterless} = this.props;
+		const {
+			className,
+			style,
+			children,
+			isGutterless
+		} = this.props;
```
nathaniellee: Can we alphabetize these while we're at it? ;)

src/components/Panel/Panel.jsx
```diff
@@ -49,22 +49,39 @@ const Panel = React.createClass(createLucidComponentDefinition({
 	},
 
 	render: function() {
-		const {className, style, children, isGutterless} = this.props;
+		const {
+			className,
+			style,
+			children,
+			isGutterless
```
nathaniellee: Does _anyone_ else think this prop name is strange? It seems to add cognitive load to give a prop a "falsey" name that would have a `true` value as opposed to having a "truthy" name like `hasGutters`.

jondlm: I agree, though we should change it on bother `Grid` and here, and technically it's a breaking change.

PR #173, opened 2016-04-06
VerticalListMenu component

jondlm:
## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [x] IE11 (Win 7)
  - [x] Edge (Win 10)


src/components/VerticalListMenu/VerticalListMenu.jsx
```diff
@@ -0,0 +1,209 @@
...
+						<li
+							key={index}
+							{...itemChildProp.passThroughs}
+							className={boundClassNames('&-Item', itemChildProp.className)}
+						>
+							<div
+								className={boundClassNames('&-Item-content', {
+									'&-Item-content-is-selected': actualIsSelected,
+								})}
+								onClick={_.partial(this.handleClickItem, index, itemChildProp)}
```
nathaniellee: Now that we're passing props through to event handlers is there a way to avoid using `partial`?

jondlm: There probably is. Let me look into it.

jondlm: So yeah, it's definitely possible but it'll take a little bit of refactoring. I'm going to leave it for now to get the ticket closed and open a perf ticket as tech debt.

src/components/VerticalListMenu/VerticalListMenu.jsx
```diff
@@ -0,0 +1,209 @@
...
+			>
+				{_.map(itemChildProps, (itemChildProp, index) => {
+					const {
+						hasExpander = false,
+					} = itemChildProp;
+
+					const itemChildrenAsArray = React.Children.toArray(itemChildProp.children);
+
+					// Was not able to get `child.Type` to work correctly, I suspect this
+					// is due to the way we wrap components with createLucidComponentDefinition
```
ogupte: `createLucidComponentDefinition` should not impact component types, also check for `child.type` (lowercase)

src/components/VerticalListMenu/VerticalListMenu.jsx
```diff
@@ -0,0 +1,209 @@
...
+					const listChildren = _.filter(itemChildrenAsArray, (child) => _.get(child, 'type.displayName', '') === 'VerticalListMenu');
```
ogupte: in the `child-component` util, we have a function called [findElementsByType](https://github.com/appnexus/lucid/blob/master/src/util/child-component.js#L87), that does this same logic but compares the actual types. `findElementsByType(children, [VerticalListMenu])`. i have a branch that does the below logic, with a function called `rejectElementByType` which is almost the same logic.

src/components/VerticalListMenu/VerticalListMenu.spec.jsx
```diff
@@ -0,0 +1,160 @@
+import _ from 'lodash';
```
ogupte: Many of our tests are organized like
```
component {
  render {...}
  props {...}
  childComponents {...}
}
``` but this file kind of breaks that pattern.

File: src/components/VerticalListMenu/VerticalListMenu.jsx (hunk 3 above)
Comment (jondlm) [inline src/components/VerticalListMenu/VerticalListMenu.jsx]: Awesome, I do remember seeing that land. I'll refactor to use that.

File: src/components/VerticalListMenu/VerticalListMenu.spec.jsx [hunk 5]
```diff
@@ -0,0 +1,160 @@
...
+		assert.equal(wrapper.find('.lucid-VerticalListMenu-Item-expander').length, 2);
+	});
+});
+
+describeWithDOM('VerticalListMenu', () => {
+	// Since onToggle uses a Button under the hood, we need to do a `mount` test
+	// to make sure it's passing the `event` object around correctly
+	it('should fire onToggle', () => {
+		const onToggle = sinon.spy();
+		const wrapper = mount(
```
ogupte: It's a good idea to call `wrapper.unmount` _after each_ test just in case the component does any cleanup on `componentWillUnmount`. Even if it doesn't do anything, it's a good habit to get into.

src/components/VerticalListMenu/VerticalListMenu.spec.jsx
```diff
@@ -0,0 +1,160 @@
...
+			<VerticalListMenu onSelect={onSelect}>
+				<VerticalListMenu.Item>One</VerticalListMenu.Item>
+				<VerticalListMenu.Item>Two</VerticalListMenu.Item>
+			</VerticalListMenu>
+		);
+
+		wrapper.find('.lucid-VerticalListMenu-Item-content').at(1).simulate('click');
+
+		assert(onSelect.called);
+		assert.equal(onSelect.args[0][0], 1, 'wrong index on the onSelect callback');
```
ogupte: i like the assertion text, we should definition keep doing more of this 👍

src/components/VerticalListMenu/examples/1.basic.jsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import VerticalListMenuStateless from '../VerticalListMenu';
+import { buildStatefulComponent } from '../../../util/state-management';
+
+const VerticalListMenu = buildStatefulComponent(VerticalListMenuStateless);
+
+export default React.createClass({
+	render() {
+		return (
+			<VerticalListMenu style={{ width: '250px' }}>
```
ogupte: width could probably just take an integer of `250` rather than the px string. They are equivalent, but we should probably favor one over the other.

src/components/VerticalListMenu/examples/2.nested-with-expander.jsx
```diff
@@ -0,0 +1,56 @@
...
+						selectedIndices={this.state.currentList === 'two' ? this.state.selectedIndices : []}
+					>
+						<VerticalListMenu.Item>Level two</VerticalListMenu.Item>
+						<VerticalListMenu.Item hasExpander={true}>
+							Level two with VerticalListMenu and lots of text. Lorem quos natus mollitia nihil quasi! Necessitatibus corporis aliquam quam laborum nesciunt quaerat. Nostrum distinctio officiis adipisci nulla unde repellat. Soluta eaque ex obcaecati molestiae provident aspernatur sit! Expedita et.
+							<VerticalListMenu
+								onSelect={_.partial(this.handleSelect, 'three')}
+								selectedIndices={this.state.currentList === 'three' ? this.state.selectedIndices : []}
+							>
+								{_.times(50, (n) => {
```
ogupte: nice, i didn't know `times` gave the iteratee the current counter number

src/components/VerticalListMenu/VerticalListMenu.reducers.js
```diff
@@ -0,0 +1,17 @@
...
+}
+
+export function onToggle(state = {}, index) {
+	const expandedIndices = state.expandedIndices || [];
+
+	return {
+		...state,
+		expandedIndices: _.xor(expandedIndices, [index]),
+	}
+}
```
ogupte: don't forget to write tests for your reducers!

jondlm: Good call!

- [ ] write tests for reducers

src/components/VerticalListMenu/examples/1.basic.jsx
jondlm: I feel like we've talked about this before. My vote is for a string with `px` since it's more explicit but I don't have strong feelings.

src/components/VerticalListMenu/VerticalListMenu.spec.jsx
jondlm: - [ ] unmount after each mounted test

src/components/VerticalListMenu/VerticalListMenu.spec.jsx
jondlm: I never noticed that before.

- [ ] rearrange tests to be more idiomatic

src/components/VerticalListMenu/VerticalListMenu.spec.jsx
jondlm: Yeah, it's nice but kind of a pain to write every time :shrug:

src/components/VerticalListMenu/VerticalListMenu.jsx
nathaniellee: Does that mean that to find "all other types" one would have to leverage something like `_.without`?

src/components/VerticalListMenu/VerticalListMenu.jsx
```diff
@@ -0,0 +1,209 @@
...
+							<div
+								className={boundClassNames('&-Item-content', {
+									'&-Item-content-is-selected': actualIsSelected,
+								})}
+								onClick={_.partial(this.handleClickItem, index, itemChildProp)}
+							>
+							<span className={boundClassNames('&-Item-content-text')}>
+								{otherChildren}
+							</span>
+								{hasExpander ?
```
nathaniellee: Something seems off with this block's indentation level... it seems like `{hasExpander ?` should align with the `</span>` above it...

src/components/VerticalListMenu/VerticalListMenu.spec.jsx
nathaniellee: Hmm... interesting... my pattern for tests has been primarily around the props and the event handlers.

src/components/VerticalListMenu/examples/1.basic.jsx
nathaniellee: The React docs themselves highlight the built-in feature of automatically converting a number value to a string value ending in `px`: https://facebook.github.io/react/tips/style-props-value-px.html

src/components/VerticalListMenu/VerticalListMenu.jsx
jondlm: Yeah, I think it's use `_.reject` actually

src/components/VerticalListMenu/VerticalListMenu.jsx
jondlm: Yep, you're right! Fixed.

PR #176, opened 2016-04-06
DragCaptureZone helper component.

nathaniellee:
## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [x] IE11 (Win 7)
  - [x] Edge (Win 10)


src/components/DragCaptureZone/DragCaptureZone.jsx
```diff
@@ -0,0 +1,148 @@
...
+	},
+
+	getDefaultProps() {
+		return {
+			onDrag: _.noop,
+			onDragEnd: _.noop,
+			onDragStart: _.noop
+		};
+	},
+
```
jondlm: Should we have an `componentWillUnmount` handle to make sure we don't have dangling event listeners if the user was dragging and it got removed from the dom for some reason? Perhaps that's overkill?

src/components/DragCaptureZone/DragCaptureZone.spec.jsx
```diff
@@ -0,0 +1,160 @@
...
+			onDragStart = sinon.spy();
+
+			mount(<DragCaptureZone onDragStart={onDragStart} />).find('div').simulate('mousedown', dragStartCoordinates);
+		});
+
+		it('calls the function passed in as the `onDragStart` prop...', () => {
+			assert(onDragStart.calledOnce);
+		});
+
+		it('...passes along a map of coordinate data as the first argument...', () => {
```
jondlm: It's a little weird to `...` tests together IMO. Each test should be a unit unto itself. It's not a big deal, and I wouldn't bother changing it for these tests, just food for thought.

src/components/DragCaptureZone/examples/interactive.jsx
```diff
@@ -0,0 +1,99 @@
+import _ from 'lodash';
```
jondlm: Love the example!

src/components/DragCaptureZone/DragCaptureZone.spec.jsx
nathaniellee: I do that mainly for readability for whoever is consuming the tests where I feel that there's a viable separation of tests. I don't feel _super_ strongly about combining these together, though.

src/components/DragCaptureZone/DragCaptureZone.jsx
nathaniellee: I don't _think_ that this would happen. While you're dragging around, it's basically a bunch of individual `mousemove` event handlers being fired so if the component were to be unmounted while you were doing that it would just unmount somewhere in between two of those `mousemove` events and you'd lose the response.

nathaniellee: Oh wait... but there _are_ the window-scoped event listeners... were _those_ the ones you were referring to? Yeah, we should definitely nix those two... good call!

PR #182, opened 2016-04-12
StickySection component

ogupte:
This component is required for Sticky Headers to be implemented for Table.

## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [x] IE11 (Win 7)
  - [x] Edge (Win 10)


danhahn:
I found that at narrow window sizes like 700px or less the sticky headers start to miss behave.  In some cases they did not stick at all in other cases they stuck but unstuck sooner than expected.  The more narrow the page the quicker it will get unstuck.

ogupte:
correct, but that is expected since the `lowerBound` prop values are constant. There is a another component which will pass in a variable value for `lowerBound` and do what you expect.

src/components/StickySection/StickySection.jsx
```diff
@@ -0,0 +1,191 @@
...
+
+		const {
+			isAboveFold,
+			containerRect,
+		} = this.state;
+
+		return (
+			<div
+				{...passthrus}
+				className={boundClassNames('&', '&-base', className)}
```
jondlm: is `-base` necessary?

ogupte: probably not

PR #197, opened 2016-04-21
Component ScrollTable

ogupte:
## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [x] IE11 (Win 7)
  - [x] Edge (Win 10)

### DIFF
--- a/src/components/ScrollTable/ScrollTable.jsx
+++ b/src/components/ScrollTable/ScrollTable.jsx
@@ -1,8 +1,6 @@
 import React from 'react';
-import ReactDOM from 'react-dom';
 import { createLucidComponentDefinition } from '../../util/component-definition';
 import { lucidClassNames } from '../../util/style-helpers';
-import { findElementsByType } from '../../util/child-component';
 import Table from '../Table/Table';
 
 const boundClassNames = lucidClassNames.bind('&-ScrollTable');
--- a/src/components/ScrollTable/ScrollTable.jsx
+++ b/src/components/ScrollTable/ScrollTable.jsx
@@ -68,12 +68,12 @@ const ScrollTable = React.createClass(createLucidComponentDefinition({
 				style={style}
 			>
 				<Table
+					{...passThroughs}
 					style={{
 						width: tableWidth,
 					}}
 					className={boundClassNames({'&-nowrap': isNowrap})}
 					hasNoBorder={true}
-					{...passThroughs}
 				>
 					{children}
 				</Table>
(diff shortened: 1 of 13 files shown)


src/components/ScrollTable/ScrollTable.jsx
```diff
@@ -0,0 +1,85 @@
...
+	},
+
+	render() {
+		const {
+			children,
+			className,
+			style,
+			tableWidth,
+			isNowrap,
+			...passThrus
```
sodiumjoe: this is spelled `passThroughs` everywhere else

src/components/Table/Table.jsx
```diff
@@ -458,10 +459,16 @@ const Table = React.createClass(createLucidComponentDefinition({
 		 * Adjusts the style of the table to have more spacing within the table cells
 		 */
 		hasExtraWhitespace: bool,
+
+		/**
+		 * render the table without borders on the outer edge
+		 */
+		hasNoBorder: bool,
```
sodiumjoe: is there a reason for this to be `hasNoBorder` with default `false` instead of `hasBorder` with default `true`?

ogupte: yes, the css class name is `has-no-border` and it actually removes existing default border styles, so the negation follows suit with that.

src/components/ScrollTable/ScrollTable.jsx
```diff
@@ -0,0 +1,85 @@
...
+				className={boundClassNames('&', className)}
+				style={style}
+			>
+				<Table
+					style={{
+						width: tableWidth,
+					}}
+					className={boundClassNames({'&-nowrap': isNowrap})}
+					hasNoBorder={true}
+					{...passThroughs}
```
jondlm: Probably want this above the other props

src/components/ScrollTable/ScrollTable.less
```diff
@@ -0,0 +1,15 @@
+.lucid-ScrollTable {
+	border: @Table-cell-border;
```
jondlm: Any time we use variables from other components, we need to hoist them up in to `variables.less`

src/components/Table/Table.jsx
```diff
@@ -230,9 +230,6 @@ const Th = React.createClass(createLucidComponentDefinition({
 			passiveWidth: width || null
 		};
 	},
-	componentDidMount() {
-		this._root = this.refs['root'];
```
jondlm: This did seem redundant to me, so 👍

PR #198, opened 2016-04-21
Proposal: redux binding utilities

sodiumjoe:
moved https://github.com/appnexus/lucid/pull/211

## PR Checklist

- [ ] finalize names

~~[x] Chrome~~
~~[ ] Firefox~~
~~[ ] Safari~~
~~[ ] IE11 (Win 7)~~
~~[ ] Edge (Win 10)~~

- [x] Unit tests written ~~(`common` at minimum)~~

~~[ ] One core team UX approval~~ n/a

+ utility functions for getting react-redux primitives from lucid
	reducers/selectors:
	+ reducer
	+ connectors (mapStateToProps, mapDispatchToProps, mergeProps)

### DIFF
--- a/src/util/redux.js
+++ b/src/util/redux.js
@@ -50,7 +50,7 @@ export function getReduxPrimitives({
 	let dispatchTree;
 
 	const reducer = createReduxReducer(reducers, initialState, rootPath);
-	const selector = reduceSelectors(selectors);
+	const selector = selectors ? reduceSelectors(selectors) : identity;
 	const rootPathSelector = state => isEmpty(rootPath) ? state : get(state, rootPath);
 	const mapStateToProps = createSelector(
 		[rootPathSelector],
--- a/src/util/redux.js
+++ b/src/util/redux.js
@@ -1,0 +1,246 @@
+import {
+	assign,
+	get,
+	identity,
+	isArray,
+	isEmpty,
+	isFunction,
+	isUndefined,
+	merge,
+	reduce,
+	slice
+} from 'lodash';
+import { createSelector } from 'reselect';
+
+/**
+ * thunk
+ *
+ * Marks a function on the reducer tree as a thunk action creator so it doesn't
+ * get incorporated into the redux reducer
+ *
+ * @return {function} with `isThunk` set to `true`
+ */
+export function thunk(fn) {
+	fn.isThunk = true;
+	return fn;
+}
+
+/**
+ * getReduxPrimitives
+ *
+ * Creates a redux reducer and connectors (inputs to redux-react's `connect`)
+ *
+ * @param {Object} param
+ * @param {Object} param.initialState - the initial state object that the reducer will return
+ * @param {Object} param.reducers - a tree of lucid reducers
+ * @param {string[]} param.rootPath - array of strings representing the path to local state in global state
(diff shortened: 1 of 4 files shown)


src/util/redux.js
```diff
@@ -0,0 +1,246 @@
+import {
```
sodiumjoe: should this just go into `src/util/state-management.js` instead?

src/util/redux.js
```diff
@@ -0,0 +1,246 @@
...
+	rootPath = [],
+	rootSelector = identity,
+	selectors
+}) {
+
+	// we need this in scope so actionCreators can refer to it
+	let dispatchTree;
+
+	const reducer = createReduxReducer(reducers, initialState, rootPath);
+	const selector = reduceSelectors(selectors);
```
sodiumjoe: naming?

I'm still leaning toward `selector` because they literally are `reselect` `selectors`.

src/util/redux.js
```diff
@@ -0,0 +1,246 @@
...
+ *
+ * @param {Object} param
+ * @param {Object} param.initialState - the initial state object that the reducer will return
+ * @param {Object} param.reducers - a tree of lucid reducers
+ * @param {string[]} param.rootPath - array of strings representing the path to local state in global state
+ * @param {Object} param.selectors - a tree of lucid selectors
+ * @return {Object} redux reducer and connectors
+ */
+
+export function getReduxPrimitives({
```
sodiumjoe: this name is not great

src/util/redux.js
sodiumjoe: how does the team feel about this desctuctured import of lodash methods? I think it makes the code below less noisy, but i can change it if nobody else likes it

ogupte: i'm fine with it as long we're consistent and change it it all the source files, so maybe save this for another PR

ogupte: if this is specific to `redux` then it belongs in it's own file, `state-management.js` is not specific to redux.

sodiumjoe: k, i'll change the lodash imports

src/util/redux.js
jondlm: Seems okay to me?

src/util/redux.spec.js
```diff
@@ -0,0 +1,197 @@
...
+				assert.deepEqual(initialState, nextState);
+			});
+		});
+
+	});
+
+	describe('connectors', () => {
+		describe('selector/mapStateToProps', () => {
+
+			const selectors = {
```
jondlm: Ahh, this is really cool!

src/util/redux.spec.js
```diff
@@ -0,0 +1,197 @@
...
+
+			describe('rootSelector', () => {
+
+				const {
+					connectors: [mapStateToProps]
+				} = getReduxPrimitives({
+					reducers,
+					initialState,
+					selectors,
+					rootSelector: state => assign({}, state, {
```
jondlm: Do you envision this extra layer of computed values to be useful for things like validation? I like the approach.

sodiumjoe: Yeah, validation was the use-case that prompted the approach.

sodiumjoe:
moved https://github.com/appnexus/lucid/pull/211

PR #202, opened 2016-04-23
Component DataTable

ogupte:
## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [x] IE11 (Win 7)
  - [x] Edge (Win 10)

### DIFF
--- a/src/components/DataTable/DataTable.jsx
+++ b/src/components/DataTable/DataTable.jsx
@@ -234,7 +234,13 @@ const DataTable = React.createClass(createLucidComponentDefinition({
 				</Thead>
 				<Tbody>
 					{_.map(data, (row, index) => (
-						<Tr isSelected={row.isSelected} isActive={row.isActive} onClick={_.partial(this.handleRowClick, index)} hasDetails={hasDetails} key={'row' + index}>
+						<Tr
+							isSelected={row.isSelected}
+							isActive={row.isActive}
+							onClick={_.partial(this.handleRowClick, index)}
+							hasDetails={hasDetails}
+							key={'row' + index}
+						>
 							{isSelectable ? (
 								<Td hasCheckbox><Checkbox isSelected={row.isSelected} onSelect={_.partial(this.handleSelect, index)} /></Td>
 							) : null}
--- a/src/components/DataTable/DataTable.jsx
+++ b/src/components/DataTable/DataTable.jsx
@@ -1,0 +1,260 @@
+import _ from 'lodash';
+import React from 'react';
+import { lucidClassNames } from '../../util/style-helpers';
+import { createLucidComponentDefinition } from '../../util/component-definition';
+import { findAllChildComponents } from '../../util/child-component';
+
+import Checkbox from '../Checkbox/Checkbox';
+import ScrollTable from '../ScrollTable/ScrollTable';
+
+const {
+	Thead,
+	Tbody,
+	Tr,
+	Th,
+	Td
+} = ScrollTable;
+
+const boundClassNames = lucidClassNames.bind('&-DataTable');
+
+const {
+	any,
+	func,
+	object,
+	string,
+	bool,
+	arrayOf,
+} = React.PropTypes;
+
+/**
+ *
+ * {"categories": ["table"], "madeFrom": ["Table","ScrollTable"]}
+ *
+ * `DataTable` provides a simple abstraction over the `Table` component to make it easier to define data-driven tables and render an array of objects.
+ */
+const DataTable = React.createClass(createLucidComponentDefinition({
+	displayName: 'DataTable',
+
+	propTypes: {
+		/**
+		 * Styles that are passed through to the root container.
+		 */
+		style: object,
+
+		/**
+		 * Class names that are appended to the defaults.
+		 */
+		className: string,
+		/**
+		 * Array of objects to be rendered in the table. Object keys match the `field` of each defined `DataTable.Column`.
+		 */
+		data: arrayOf(object),
+		/**
+		 * Adjusts the style of the table to have more spacing within the table cells
+		 */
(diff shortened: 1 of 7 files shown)


src/components/DataTable/DataTable.jsx
```diff
@@ -0,0 +1,260 @@
...
+	},
+
+	render() {
+
+		const {
+			className,
+			data,
+			hasDetails,
+			isSelectable,
+			...passThrus
```
sodiumjoe: `passThroughs`

src/components/DataTable/DataTable.spec.jsx
```diff
@@ -0,0 +1,498 @@
...
+						<Column field='id' title='ID'/>
+						<Column field='first_name' title='First'/>
+						<Column field='last_name' title='Last'/>
+						<Column field='email' title='Email'/>
+						<Column field='occupation' title='Occupation'/>
+					</DataTable>
+				);
+
+				// select the rows of the rendered table head
+				//const headTrsWrapper = wrapper
```
sodiumjoe: delete?

src/components/DataTable/examples/1.basic.jsx
```diff
@@ -0,0 +1,122 @@
...
+	{
+		'id': '10',
+		'first_name': 'William',
+		'last_name': 'Shakespeare',
+		'email': 'wshakespear@example.com',
+		'occupation': 'Playwright',
+	},
+];
+
+export default React.createClass({
```
sodiumjoe: this is sweet

src/components/DataTable/DataTable.jsx
```diff
@@ -0,0 +1,260 @@
...
+	func,
+	object,
+	string,
+	bool,
+	arrayOf,
+} = React.PropTypes;
+
+/**
+ *
+ * {"categories": ["table"], "madeFrom": ["Table","ScrollTable"]}
```
jondlm: I think this should be `Checkbox` and `ScrollTable`

jondlm:
Seeing a minor issue with the caret on Chrome at least.

eh

src/components/DataTable/DataTable.jsx
ogupte: i can update to this, but i still think Table is useful here since they need to reference Table.Th to see what props are available to `Column` and `ColumnGroup`. Also Checkbox, while it's true it's make from it, doesn't expose any of it's props thru this abstraction, so it may not be very useful to the reader of the docs.

ogupte:
that is actually a bug in the `Table` component which i will create an issue for right now.

src/components/DataTable/DataTable.jsx
jondlm: Yeah, I hear you. I really wanted to keep `madeFrom` based strictly on composition so we could draw a dependency graph of our library. It might be overkill, but we could have a `links` sections that gives people the most useful links to other components.

src/components/DataTable/DataTable.jsx
```diff
@@ -0,0 +1,260 @@
...
+								{props.title || props.children}
+							</Th>
+						))}
+					</Tr>
+					{hasGroupedColumns ? (
+						<Tr>
+							{_.reduce(flattenedColumns, (acc, { props: columnProps, columnGroupProps }, index) => acc.concat(_.isNull(columnGroupProps) ? [] : [(
+								<Th
+									{..._.omit(columnProps, ['field', 'children', 'width', 'title'])}
+									onClick={DataTable.shouldColumnHandleSort(columnProps) ? _.partial(this.handleSort, columnProps.field) : null}
```
jondlm: It's not a big deal right for now, but I can imagine partials causing perf problems down the road. If/when we run into issues we should probably drop the `columnProps.field` as a prop on the `Th` and destructure it out of the callback when fired.

src/components/DataTable/DataTable.jsx
```diff
@@ -0,0 +1,260 @@
...
+								>
+									{columnProps.title || columnProps.children}
+								</Th>
+							)]), [])}
+						</Tr>
+					) : null}
+				</Thead>
+				<Tbody>
+					{_.map(data, (row, index) => (
+						<Tr isSelected={row.isSelected} isActive={row.isActive} onClick={_.partial(this.handleRowClick, index)} hasDetails={hasDetails} key={'row' + index}>
```
jondlm: These long lines are hard to read. Again, it's okay for now but let's drop them in the future:

```
<Tr
	isSelected={row.isSelected}
	isActive={row.isActive}
	onClick={_.partial(this.handleRowClick, index)}
	hasDetails={hasDetails}
	key={'row' + index}
>
```

src/components/DataTable/DataTable.spec.jsx
```diff
@@ -0,0 +1,486 @@
...
+				.find(ScrollTable).shallow();
+
+			assert(wrapper.is(ScrollTable), 'must render a ScrollTable');
+			assert(wrapper.is('.lucid-DataTable'), 'must have the component className');
+			assert.equal(scrollTableWrapper.find(ScrollTable.Thead).length, 1, 'must contain a Thead');
+			assert.equal(scrollTableWrapper.find(ScrollTable.Tbody).length, 1, 'must contain a Tbody');
+		});
+	});
+
+	describe('props', () => {
```
jondlm: I love the readability of these tests, nice work!

src/components/DataTable/examples/1.basic.jsx
jondlm: Yeah, I love the API. I'm considering doing something similar for charts with series.

src/components/DataTable/DataTable.jsx
ogupte: ok i'm on board

PR #206, opened 2016-04-28
Appnexus/anxr 372 export default stateful

sodiumjoe:
CHANGED TARGET: https://github.com/appnexus/lucid/pull/206

There's a lot of noise from fixing all the lint errors, so it's probably easiest to look at the commits individually.

I'm not sure what semver label to give this. I think it's not really a breaking change, so... minor?

## PR Checklist

  - ~~[ ] Chrome~~
  - ~~[ ] Firefox~~
  - ~~[ ] Safari~~
  - ~~[ ] IE11 (Win 7)~~
  - ~~[ ] Edge (Win 10)~~

- [x] https://github.com/appnexus/lucid/pull/206#discussion_r61453813
- [x] https://github.com/appnexus/lucid/pull/206#issuecomment-215273896
- [x] run buildStateful deprecation warning only once

### DIFF
--- a/src/components/Autocomplete/Autocomplete.jsx
+++ b/src/components/Autocomplete/Autocomplete.jsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import _ from 'lodash';
 import { lucidClassNames } from '../../util/style-helpers';
-import { createLucidComponentDefinition } from '../../util/component-definition';
+import { createClass } from '../../util/component-definition';
 import * as reducers from './Autocomplete.reducers';
 import * as KEYCODE from '../../constants/key-code';
 import DropMenu from '../DropMenu/DropMenu';
@@ -25,7 +25,7 @@ const {
  * A text input with suggested values displayed in an attached menu.
  */
 
-const Autocomplete = React.createClass(createLucidComponentDefinition({
+const Autocomplete = createClass({
 	displayName: 'Autocomplete',
 
 	reducers,
@@ -275,7 +275,7 @@ const Autocomplete = React.createClass(createLucidComponentDefinition({
 			</DropMenu>
 		);
 	}
-}));
+});
 
 function partitionText(text, pattern, length) {
 	const index = text.search(pattern);
--- a/src/components/Badge/Badge.jsx
+++ b/src/components/Badge/Badge.jsx
@@ -7,8 +7,6 @@ const boundClassNames = lucidClassNames.bind('&-Badge');
 const {
 	node,
 	string,
-	oneOfType,
-	arrayOf,
 } = React.PropTypes;
 
 /**
@@ -29,10 +27,7 @@ const Badge = createClass({
 		/**
 		 * any valid React children
 		 */
-		children: oneOfType([
-			node,
-			arrayOf(node)
-		]),
+		children: node,
 	},
 
 	render() {
--- a/src/components/Badge/Badge.jsx
+++ b/src/components/Badge/Badge.jsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import { lucidClassNames } from '../../util/style-helpers';
+import { createClass } from '../../util/component-definition';
 
 const boundClassNames = lucidClassNames.bind('&-Badge');
 
@@ -18,7 +19,8 @@ const {
  * element(s). Do not wrap existing `Icon`s in a badge, rather add the
  * `isBadge` prop to any Icon component to turn it into a badge.
  */
-const Badge = React.createClass({
+const Badge = createClass({
+	displayName: 'Badge',
 	propTypes: {
 		/**
 		 * class names that are appended to the defaults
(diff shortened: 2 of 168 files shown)


src/components/Autocomplete/Autocomplete.jsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import _ from 'lodash';
 import { lucidClassNames } from '../../util/style-helpers';
-import { createLucidComponentDefinition } from '../../util/component-definition';
+import { createClass } from '../../util/component-definition';
```
sodiumjoe: this wasn't strictly necessary for this, but I had already done the work of changing them all to use `createClass` and it sets us up for some refactoring work on child component definition and `createLucidComponentDefinition`, so i left it

src/index.js
```diff
@@ -1,50 +1,65 @@
-import Autocomplete from './components/Autocomplete/Autocomplete';
+import { buildStatefulComponent } from './util/state-management';
```
sodiumjoe: this kinda sucks and i'm halfway tempted to change this file to a common js export syntax so we can dynamically generate the exports... at least until we have some static analysis use-case for using es6 exports

sodiumjoe:
this is gonna be painful, but i think we should also switch all the examples to pull from `src/index.js` so that they use the hybrid components.

jondlm:
I agree that we should fix the examples to pull from index.

src/components/Badge/Badge.jsx
```diff
@@ -27,10 +27,7 @@ const Badge = React.createClass({
 		/**
 		 * any valid React children
 		 */
-		children: oneOfType([
-			node,
-			arrayOf(node)
-		]),
+		children: node,
```
jondlm: Bless your heart good sir!

src/components/TextField/TextField.jsx
```diff
@@ -77,10 +77,17 @@ const TextField = React.createClass(createLucidComponentDefinition({
...
+		 * Signature: `(value, { event, props }) => {}`
+		 */
+		onKeyDown: func,
+
+		/**
 		 * Fires an event when the user hits "enter" from the TextField. You
 		 * shouldn't use it if you're using `isMultiLine`.
 		 *
-		 * Signature: `(value, { event, props }) => {}`
+		 * Signature: `(event) => {}`
```
jondlm: Was this an accident?

https://github.com/appnexus/lucid/blob/appnexus/ANXR-372-export-default-stateful/src/components/TextField/TextField.jsx#L203

src/index.js
jondlm: I'm okay with it. We should probably have a common unit test that enforces the convention though.

src/util/state-management.js
```diff
@@ -121,3 +129,10 @@ export function buildStatefulComponent(baseComponent, opts) {
 		}
 	});
 }
+
+export function buildStatefulComponent(...args) {
```
jondlm: Nice!

src/util/state-management.js
```diff
@@ -82,7 +82,13 @@ function overwriteArrays (objValue, srcValue) {
 	}
 }
 
-export function buildStatefulComponent(baseComponent, opts) {
+export function buildHybridComponent(baseComponent, opts) {
+
+	if (baseComponent._isLucidHybridComponent && process.NODE_ENV !== 'production') {
```
jondlm: Not sure about that `production` check. Shouldn't it be this instead?

```
if (baseComponent._isLucidHybridComponent) {
	if ()process.NODE_ENV !== 'production') {
		console.warn('Input component is already a Lucid hybdrid component. Lucid exports hybrid components by default. To access the dumb components, use the -Dumb suffix, e.g. "ComponentDumb"');
	}

	return baseComponent;
}
```

src/components/TextField/TextField.jsx
sodiumjoe: what do you mean? keydown appears to only send the event: https://github.com/appnexus/lucid/blob/appnexus/ANXR-372-export-default-stateful/src/components/TextField/TextField.jsx#L199

tho we should probably change that so it's consistent with the others

src/util/state-management.js
sodiumjoe: yep, you're right

sodiumjoe: and, actually, about that `production` check, i didn't find anywhere else where we're doing that. Is that how we want to do it?

src/components/TextField/TextField.jsx
jondlm: I think you swapped the signature for `onKeyDown` and `onSubmit`

jondlm: I wanted to leave `onKeyDown` vanilla since it wasn't really supposed to be part of the API of this component. I just wanted to make sure it would work as expected if the consumer was trying to `passThrough` a key down handler.

sodiumjoe: uh, derp you're right

sodiumjoe: not sure why, but i keep thinking the trailing comment applies instead of the preceding one

PR #207, opened 2016-04-28
Change Tabs onSelect signature

jacklenehan:
## PR Checklist


jondlm:
I think technically this is a breaking change since a callback signature is changing. However, it's a very low risk breaking change IMO.

sodiumjoe:
maybe we should save it for `1.0`

PR #209, opened 2016-04-29
Appnexus/anxr 372 export default stateful

sodiumjoe:
changed PR target of https://github.com/appnexus/lucid/pull/206 to `release/1.0.0`

## PR Checklist

  - ~~[ ] Chrome~~
  - ~~[ ] Firefox~~
  - ~~[ ] Safari~~
  - ~~[ ] IE11 (Win 7)~~
  - ~~[ ] Edge (Win 10)~~

- [x] https://github.com/appnexus/lucid/pull/206#discussion_r61453813
- [x] https://github.com/appnexus/lucid/pull/206#issuecomment-215273896
- [x] run buildStateful deprecation warning only once
- [ ] finalize "dumb" component suffix

### DIFF
--- a/src/components/SingleSelect/SingleSelect.jsx
+++ b/src/components/SingleSelect/SingleSelect.jsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import _ from 'lodash';
 import { lucidClassNames } from '../../util/style-helpers';
-import { createLucidComponentDefinition } from '../../util/component-definition';
+import { createClass } from '../../util/component-definition';
 import * as reducers from './SingleSelect.reducers';
 import DropMenu from '../DropMenu/DropMenu';
 import CaretIcon from '../Icon/CaretIcon/CaretIcon';
@@ -27,7 +27,7 @@ const {
  * Supports option groups with and without labels.
  */
 
-const SingleSelect = React.createClass(createLucidComponentDefinition({
+const SingleSelect = createClass({
 	displayName: 'SingleSelect',
 
 	reducers,
@@ -189,6 +189,6 @@ const SingleSelect = React.createClass(createLucidComponentDefinition({
 			</DropMenu>
 		);
 	}
-}));
+});
 
 export default SingleSelect;
--- a/src/components/Table/Table.jsx
+++ b/src/components/Table/Table.jsx
@@ -1,7 +1,7 @@
 import _ from 'lodash';
 import React from 'react';
 import { lucidClassNames } from '../../util/style-helpers';
-import { createLucidComponentDefinition } from '../../util/component-definition';
+import { createClass } from '../../util/component-definition';
 import { findElementsByType } from '../../util/child-component';
 import CaretIcon from '../Icon/CaretIcon/CaretIcon';
 import DragCaptureZone from '../DragCaptureZone/DragCaptureZone';
@@ -21,7 +21,7 @@ const {
  *
  * Any child `<Tr>` will have `isHeader` set to `true` unless otherwise specified.
  */
-const Thead = React.createClass(createLucidComponentDefinition({
+const Thead = createClass({
 	render() {
 		const {
 			children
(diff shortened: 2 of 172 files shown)


src/components/SingleSelect/SingleSelect.jsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import _ from 'lodash';
 import { lucidClassNames } from '../../util/style-helpers';
-import { createLucidComponentDefinition } from '../../util/component-definition';
+import { createClass } from '../../util/component-definition';
 import * as reducers from './SingleSelect.reducers';
 import DropMenu from '../DropMenu/DropMenu';
 import CaretIcon from '../Icon/CaretIcon/CaretIcon';
```
ogupte: are we going to import DropMenu and CarentIcon from `../../../index`?

src/components/Table/Table.jsx
```diff
@@ -1,27 +1,40 @@
 import _ from 'lodash';
 import React from 'react';
 import { lucidClassNames } from '../../util/style-helpers';
-import { createLucidComponentDefinition } from '../../util/component-definition';
+import { createClass } from '../../util/component-definition';
 import { findElementsByType } from '../../util/child-component';
 import CaretIcon from '../Icon/CaretIcon/CaretIcon';
 import DragCaptureZone from '../DragCaptureZone/DragCaptureZone';
```
ogupte: import CaretIcon, DragCaptureZone from `../../../index` ?

ogupte:
i feel like this change makes the code more confusing overall, some components are being imported by other from `index` some are directly imported, `there are dumb versions of some components, but not others, are we importing the dump or not-dumb versions of components from `index` what if there was no _dumb_ version index which one are we importing then?

I feel that this could be more consistent and simple if we were to have `buildStatefulComponent` devolve in to the identity function if there are no reducers/selectors, and then _always_ hang a `.Dumb` static property off the returned component. then run it thru _all_ exported components.

src/util/util.js
```diff
@@ -0,0 +1,74 @@
+import _ from 'lodash';
```
ogupte: this file should be named `util/environment.js` or something rather than just the generic `util.js`

ogupte: also this file is missing tests

src/util/state-management.js
```diff
@@ -82,7 +86,18 @@ function overwriteArrays (objValue, srcValue) {
 	}
 }
 
-export function buildStatefulComponent(baseComponent, opts) {
+export function buildHybridComponent(baseComponent, opts) {
```
ogupte: there's no tests for this function.

src/components/SingleSelect/SingleSelect.jsx
sodiumjoe: no, that's the public hybrid version. Internally, we compose using dumb components. Right?

src/util/state-management.js
```diff
@@ -1,5 +1,9 @@
 import React from 'react';
 import _ from 'lodash';
+import { logger } from './util';
+
+let hasLoggedBuildStateComponentWarning = false;
+const alreadyWarnedHybrid = {};
```
ogupte: this log-once type logic should be part of `logger`!

sodiumjoe:
I agree that it's confusing that there are dumb versions of some components and not others. There's not really a good solution to this, though, imo. Even if we run everything through buildHybrid, that means some components have a `hybrid` version but no reducers, and also a dumb version that, to a consumer, is identical. I think it's confusing either way we fall on that one.

on `import`ing from `../../index`, we only do that in the examples, because the examples use the library like a consumer would. Components should only ever use direct references to components (i.e. the dumb versions).

src/util/util.js
sodiumjoe: fixed

src/util/state-management.js
sodiumjoe: fixed

src/util/state-management.js
sodiumjoe: fixed

PR #211, opened 2016-04-29
Redux binding utils

sodiumjoe:
re-targeted https://github.com/appnexus/lucid/pull/198

## PR Checklist

- [x] finalize names

~~[x] Chrome~~
~~[ ] Firefox~~
~~[ ] Safari~~
~~[ ] IE11 (Win 7)~~
~~[ ] Edge (Win 10)~~

- [x] Unit tests written ~~(`common` at minimum)~~

~~[ ] One core team UX approval~~ n/a

+ utility functions for getting react-redux primitives from lucid
	reducers/selectors:
	+ reducer
	+ connectors (mapStateToProps, mapDispatchToProps, mergeProps)

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -48,13 +48,15 @@ import * as componentDefinition from './util/component-definition';
 import * as domHelpers from './util/dom-helpers';
 import * as stateManagement from './util/state-management';
 import * as styleHelpers from './util/style-helpers';
+import * as redux from './util/redux';
 
 export {
 	childComponent,
 	componentDefinition,
 	domHelpers,
+	redux,
 	stateManagement,
-	styleHelpers,
+	styleHelpers
 };
 
 export {
@@ -101,5 +103,5 @@ export {
 	TextField,
 	ValidatedTextField,
 	Validation,
-	WarningIcon,
+	WarningIcon
 };
--- a/src/util/redux.js
+++ b/src/util/redux.js
@@ -23,6 +23,7 @@ export function thunk(fn) {
  * @param {Object} param.initialState - the initial state object that the reducer will return
  * @param {Object} param.reducers - a tree of lucid reducers
  * @param {string[]} param.rootPath - array of strings representing the path to local state in global state
+ * @param {function} param.rootSelector - a top-level selector which takes as input state that has run through every selector in param.selectors
  * @param {Object} param.selectors - a tree of lucid selectors
  * @return {Object} redux reducer and connectors
  */
@@ -181,6 +182,7 @@ function createReducerFromReducerTree(reduxReducerTree, initialState) {
  * Generates a redux reducer from a tree of lucid reducers
  * @param {Object} reducers - a tree of lucid reducers
  * @param {Object} initialState - the initial state object that the reducer will return
+ * @param {string[]} rootPath - array of strings representing the path to part of global state this reducer applies to
  * @return {function} the redux reducer
  */
 function createReduxReducer(reducers, initialState, rootPath) {
--- a/src/util/redux.js
+++ b/src/util/redux.js
@@ -50,7 +50,7 @@ export function getReduxPrimitives({
 	let dispatchTree;
 
 	const reducer = createReduxReducer(reducers, initialState, rootPath);
-	const selector = reduceSelectors(selectors);
+	const selector = selectors ? reduceSelectors(selectors) : identity;
 	const rootPathSelector = state => isEmpty(rootPath) ? state : get(state, rootPath);
 	const mapStateToProps = createSelector(
 		[rootPathSelector],
--- a/src/util/redux.js
+++ b/src/util/redux.js
@@ -1,0 +1,246 @@
+import {
+	assign,
+	get,
+	identity,
+	isArray,
+	isEmpty,
+	isFunction,
+	isUndefined,
+	merge,
+	reduce,
+	slice
+} from 'lodash';
+import { createSelector } from 'reselect';
+
+/**
+ * thunk
+ *
+ * Marks a function on the reducer tree as a thunk action creator so it doesn't
+ * get incorporated into the redux reducer
+ *
+ * @return {function} with `isThunk` set to `true`
+ */
+export function thunk(fn) {
+	fn.isThunk = true;
+	return fn;
+}
+
+/**
+ * getReduxPrimitives
+ *
+ * Creates a redux reducer and connectors (inputs to redux-react's `connect`)
+ *
+ * @param {Object} param
+ * @param {Object} param.initialState - the initial state object that the reducer will return
+ * @param {Object} param.reducers - a tree of lucid reducers
+ * @param {string[]} param.rootPath - array of strings representing the path to local state in global state
+ * @param {Object} param.selectors - a tree of lucid selectors
+ * @return {Object} redux reducer and connectors
+ */
+
+export function getReduxPrimitives({
+	initialState,
+	reducers,
+	rootPath = [],
+	rootSelector = identity,
+	selectors
+}) {
+
+	// we need this in scope so actionCreators can refer to it
+	let dispatchTree;
+
+	const reducer = createReduxReducer(reducers, initialState, rootPath);
+	const selector = reduceSelectors(selectors);
+	const rootPathSelector = state => isEmpty(rootPath) ? state : get(state, rootPath);
+	const mapStateToProps = createSelector(
+		[rootPathSelector],
+		root => rootSelector(selector(root))
+	);
+	const mapDispatchToProps = dispatch => getDispatchTree(reducers, rootPath, dispatch);
+
+	return {
+		reducer,
+		connectors: [
+			mapStateToProps,
+			mapDispatchToProps,
+			mergeProps
+		]
+	};
+
+	/**
+	 * createActionCreator
+	 *
+	 * @param {function} node - a node in the the reducer tree, either a reducer or a thunk
+	 * @param {string[]} path - the path to the reducer in the reducer tree
+	 * @param {string[]} rootPath - array of strings representing the path to local state in global state
+	 * @return {function} action creator that returns either an action or a thunk
+	 */
+	function createActionCreator(node, path, rootPath) {
+		if (node.isThunk) {
+			return function thunk(...args) {
+				return function thunkInner(dispatch, getState) {
+					const localPath = slice(path, rootPath.length, -1);
+					const localDispatchTree = isEmpty(localPath) ? dispatchTree : get(dispatchTree, localPath);
+					const getLocalState = isEmpty(rootPath) ? getState : () => get(getState(), slice(path, 0, -1));
+					return node(...args)(localDispatchTree, getLocalState, dispatch, getState);
+				};
+			};
+		}
+		return function actionCreator(payload, ...meta) {
+			return {
+				type: path.join('.'),
+				payload,
+				meta
+			};
+		};
+	}
+
+	/**
+	 * createActionCreatorTree
+	 *
+	 * Walks the reducer tree and generates a tree of action creators that correspond to each reducer
+	 * @param {Object} reducers - a tree of lucid reducers
+	 * @param {string[]} rootPath - array of strings representing the path to local state in global state
+	 * @returns {Object} action creator tree
+	 */
+	function createActionCreatorTree(reducers, rootPath) {
+		return (function recur(reducers, path = []) {
+			return reduce(reducers, (memo, node, key) => {
+				const currentPath = path.concat(key);
+				return assign(memo, {
+					[key]: isFunction(node) ?
+						createActionCreator(node, currentPath, rootPath) :
+						recur(node, currentPath)
+				});
+			}, {});
+		})(reducers, rootPath);
+	}
+
+	/**
+	 * getDispatchTree
+	 *
+	 * Walks the reducer tree and generates an action creator tree, then binds dispatch to each node
+	 * @param {Object} reducers - a tree of lucid reducers
+	 * @param {string[]} rootPath - array of strings representing the path to local state in global state
+	 * @param {function} dispatch - the redux store's `dispatch` function
+	 */
+	function getDispatchTree(reducers, rootPath, dispatch) {
+		const actionCreatorTree = createActionCreatorTree(reducers, rootPath);
+		dispatchTree = bindActionCreatorTree(actionCreatorTree, dispatch);
+		return dispatchTree;
+	}
+
+}
+
+/**
+ * createReduxReducerTree
+ *
+ * Walks the reducer tree and generates a tree of redux reducers, converting the
+ * signature from `(state, payload) => state` to `(state, action) => state`
+ * @param {Object} reducers - a tree of lucid reducers
+ * @param {string[]} path - array of strings representing the path to the reducer
+ * @return {Object} redux reducer tree
+ */
+function createReduxReducerTree(reducers, path = []) {
+	return reduce(reducers, (memo, node, key) => {
+		// filter out thunks from the reducer tree
+		if (node.isThunk) {
+			return memo;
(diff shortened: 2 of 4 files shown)


src/index.js
```diff
@@ -48,13 +48,15 @@ import * as componentDefinition from './util/component-definition';
...
 import * as styleHelpers from './util/style-helpers';
+import * as redux from './util/redux';
 
 export {
 	childComponent,
 	componentDefinition,
 	domHelpers,
+	redux,
 	stateManagement,
-	styleHelpers,
```
ogupte: we should keep these commas

src/util/redux.js
```diff
@@ -0,0 +1,235 @@
...
+import { createSelector } from 'reselect';
+
+/**
+ * thunk
+ *
+ * Marks a function on the reducer tree as a thunk action creator so it doesn't
+ * get incorporated into the redux reducer
+ *
+ * @return {function} with `isThunk` set to `true`
+ */
```
ogupte: this is brilliant

src/util/redux.js
```diff
@@ -0,0 +1,235 @@
...
+ *
+ * @param {Object} param
+ * @param {Object} param.initialState - the initial state object that the reducer will return
+ * @param {Object} param.reducers - a tree of lucid reducers
+ * @param {string[]} param.rootPath - array of strings representing the path to local state in global state
+ * @param {Object} param.selectors - a tree of lucid selectors
+ * @return {Object} redux reducer and connectors
+ */
+
+export function getReduxPrimitives({
```
ogupte: i feel this function needs to be split up:
i want to be able to do 
``` connect(...getConnectFuncs(...))``` or something in my component, and then ```createStore(getReduxReducer(...))``` separately

File: src/util/redux.js [hunk 4]
```diff
@@ -0,0 +1,235 @@
...
+	 * @param {string[]} rootPath - array of strings representing the path to local state in global state
+	 * @return {function} action creator that returns either an action or a thunk
+	 */
+	function createActionCreator(node, path, rootPath) {
+		if (node.isThunk) {
+			return function thunk(...args) {
+				return function thunkInner(dispatch, getState) {
+					const localPath = _.slice(path, rootPath.length, -1);
+					const localDispatchTree = _.isEmpty(localPath) ? dispatchTree : _.get(dispatchTree, localPath);
+					const getLocalState = _.isEmpty(rootPath) ? getState : () => _.get(getState(), _.slice(path, 0, -1));
```
ogupte: i think it's more readable to use `_.dropRight(path, 1)` or `_.take(path, path.length - 1)` instead of the `slice`

src/util/redux.js
```diff
@@ -0,0 +1,235 @@
...
+	 * Walks the reducer tree and generates a tree of action creators that correspond to each reducer
+	 * @param {Object} reducers - a tree of lucid reducers
+	 * @param {string[]} rootPath - array of strings representing the path to local state in global state
+	 * @returns {Object} action creator tree
+	 */
+	function createActionCreatorTree(reducers, rootPath) {
+		return (function recur(reducers, path = []) {
+			return _.reduce(reducers, (memo, node, key) => {
+				const currentPath = path.concat(key);
+				return _.assign(memo, {
```
ogupte:
```return {
  ...memo,
  [key]: stuff,
}```

File: src/util/redux.js [hunk 6]
```diff
@@ -0,0 +1,235 @@
...
+	/**
+	 * createActionCreatorTree
+	 *
+	 * Walks the reducer tree and generates a tree of action creators that correspond to each reducer
+	 * @param {Object} reducers - a tree of lucid reducers
+	 * @param {string[]} rootPath - array of strings representing the path to local state in global state
+	 * @returns {Object} action creator tree
+	 */
+	function createActionCreatorTree(reducers, rootPath) {
+		return (function recur(reducers, path = []) {
```
ogupte: i don't think this `recur` function is needed, can't u just give `rootPath` a default value of `=[]` and make the recursive call on that?

src/util/redux.js
```diff
@@ -0,0 +1,235 @@
...
+	 * getDispatchTree
+	 *
+	 * Walks the reducer tree and generates an action creator tree, then binds dispatch to each node
+	 * @param {Object} reducers - a tree of lucid reducers
+	 * @param {string[]} rootPath - array of strings representing the path to local state in global state
+	 * @param {function} dispatch - the redux store's `dispatch` function
+	 */
+	function getDispatchTree(reducers, rootPath, dispatch) {
+		const actionCreatorTree = createActionCreatorTree(reducers, rootPath);
+		dispatchTree = bindActionCreatorTree(actionCreatorTree, dispatch);
```
ogupte: `return` this line instead of setting it to an unscoped variable.

src/util/redux.js
```diff
@@ -0,0 +1,235 @@
...
+ * @return {Object} redux reducer tree
+ */
+function createReduxReducerTree(reducers, path = []) {
+	return _.reduce(reducers, (memo, node, key) => {
+		// filter out thunks from the reducer tree
+		if (node.isThunk) {
+			return memo;
+		}
+		const currentPath = path.concat(key);
+		return _.assign(memo, {
```
ogupte:
```
return {
  ...memo,
  [key]: stuff,
};
```

src/util/redux.js
```diff
@@ -0,0 +1,235 @@
...
+ * @param {Object} initialState - the initial state object that the reducer will return
+ * @return {function} the redux reducer
+ */
+function createReducerFromReducerTree(reduxReducerTree, initialState) {
+	return function reduxReducer(state, action) {
+		if (_.isUndefined(state)) {
+			return initialState;
+		}
+		return _.reduce(reduxReducerTree, (state, node, key) => {
+			return _.assign({}, state, _.isFunction(node) ? node(state, action) : {
```
ogupte: can use spread instead of `assign`

src/util/redux.js
```diff
@@ -0,0 +1,235 @@
...
+		}, state);
+	};
+}
+
+/**
+ * createReduxReducer
+ *
+ * Generates a redux reducer from a tree of lucid reducers
+ * @param {Object} reducers - a tree of lucid reducers
+ * @param {Object} initialState - the initial state object that the reducer will return
```
ogupte: `rootPath` is missing from the jsdoc

src/util/redux.js
```diff
@@ -0,0 +1,235 @@
...
+	const reducerTree = createReduxReducerTree(reducers, rootPath);
+	return createReducerFromReducerTree(reducerTree, initialState);
+}
+
+/**
+ * bindActionCreatorTree
+ *
+ * Binds redux store.dispatch to actionCreators in a tree
+ * @param {Object} actionCreatorTree - a tree of redux action creator functions
+ * @param {function} dispatch - the redux store's `dispatch` function
```
ogupte: `path` missing from jsdoc

src/index.js
sodiumjoe: i think i'll make a separate PR with the lint rule and repo-wide fixes

src/util/redux.js
```diff
@@ -0,0 +1,235 @@
...
+/**
+ * getReduxPrimitives
+ *
+ * Creates a redux reducer and connectors (inputs to redux-react's `connect`)
+ *
+ * @param {Object} param
+ * @param {Object} param.initialState - the initial state object that the reducer will return
+ * @param {Object} param.reducers - a tree of lucid reducers
+ * @param {string[]} param.rootPath - array of strings representing the path to local state in global state
+ * @param {Object} param.selectors - a tree of lucid selectors
```
ogupte: `param. rootSelector` missing from jsdoc

src/util/redux.js
sodiumjoe: we need the outer scope binding because we use it in the `actionCreators`, see: https://github.com/appnexus/lucid/pull/211/files#diff-2a55f0d983e368ae3297e2a3774af4e0R72

src/util/redux.js
sodiumjoe: we're using object spread? My understanding is that it's not well-specced yet, so it's cool in jsx, but not yet js.

src/util/redux.js
```diff
@@ -0,0 +1,235 @@
...
+/**
+ * bindActionCreatorTree
+ *
+ * Binds redux store.dispatch to actionCreators in a tree
+ * @param {Object} actionCreatorTree - a tree of redux action creator functions
+ * @param {function} dispatch - the redux store's `dispatch` function
+ * @param {string[]} path - array of strings representing the path to the action creator
+ */
+function bindActionCreatorTree(actionCreatorTree, dispatch, path = []) {
+	return _.reduce(actionCreatorTree, (memo, node, key) => _.assign(memo, {
```
ogupte: can use spread operator instead of `_.assign`.

src/util/redux.js
```diff
@@ -0,0 +1,235 @@
...
+
+/**
+ * reduceSelectors
+ *
+ * Generates a root selector from a tree of selectors
+ * @param {Object} selectors - a tree of selectors
+ * @returns {function} root selector that when called with state, calls each of
+ * the selectors in the tree with the state local to that selector
+ */
+// @TODO: optimize to create and reuse selectors?
```
ogupte: should this TODO stay in the code?

sodiumjoe: i'll move it into a ticket

src/util/redux.js
sodiumjoe: need it for the explicit reference to `rootPath` that gets passed to `createActionCreator` regardless of depth in the tree. Could change the signature to `createActionCreatorTree(reducer, path = [], rootPath = [])` instead, but this seemed cleaner

src/util/redux.js
ogupte: all our reducers are already using it.

src/util/redux.js
ogupte: where is `dispatchTree` defined?

src/util/redux.js
ogupte: looks cleaner the other way: `createActionCreatorTree(reducer, path = [], rootPath = [])` imo

src/util/redux.js
sodiumjoe: `getReduxPrimitives` function scope: https://github.com/appnexus/lucid/pull/211/files#diff-2a55f0d983e368ae3297e2a3774af4e0R39

This is also the reason why a bunch of these functions are defined inside `getReduxPrimitives` instead of top level. Not ideal, but I couldn't figure out another way. Very open to suggestions.

src/util/redux.js
sodiumjoe: o damn ok

src/util/redux.spec.js
```diff
@@ -0,0 +1,197 @@
...
+	isFunction,
+	upperCase
+} from 'lodash';
+
+import {
+	thunk,
+	getReduxPrimitives
+} from './redux';
+
+describe('getReduxPrimitives', () => {
```
ogupte: we should group all these in a `describe('redux utils')` a module-level grouping.

src/util/redux.spec.js
```diff
@@ -0,0 +1,197 @@
...
+	assign,
+	isFunction,
+	upperCase
+} from 'lodash';
+
+import {
+	thunk,
+	getReduxPrimitives
+} from './redux';
+
```
ogupte: need a `describe` for each of the exported functions in `utils/redux.js` i only see a describe for `getReduxPrimitives`

ogupte: also they all should be grouped in a top-level describe that labels the whole module. E.g. `describe('redux utils')`

src/util/redux.spec.js
```diff
@@ -0,0 +1,197 @@
...
+		}
+	};
+
+	describe('reducer', () => {
+
+		const { reducer } = getReduxPrimitives({ reducers, initialState });
+
+		it('should return initialState on unmatched type', () => {
+			const action = { type: 'UNKNOWN' };
+			assert.deepEqual(reducer(initialState, action), initialState);
```
ogupte: add descriptions for each `assert` call (last param)

src/util/redux.spec.js
```diff
@@ -0,0 +1,197 @@
...
+				assert.deepEqual(dispatchedAction, {
+					type: 'foo.onChange',
+					payload: 'bar',
+					meta: ['baz']
+				});
+			});
+
+			describe('thunks', () => {
+
+				beforeEach(() => dispatchTree.foo.asyncOperation('qux'));
```
ogupte: i know this always works, but since it's techinically asyc, please call `done` at the end of the function.

src/util/redux.spec.js
```diff
@@ -0,0 +1,197 @@
...
+
+				it('should dispatch a thunk', () => {
+					const dispatchedThunk = mockDispatch.getCall(0).args[0];
+					assert(isFunction(dispatchedThunk))
+				});
+
+				it('should dispatch the correct action', () => {
+					const dispatchedAction = mockDispatch.getCall(1).args[0];
```
ogupte: would this still pass if it was an `it.only` or is the test order-dependent?

sodiumjoe: it would work, `dispatchTree.foo.asyncOperation` calls them in order

src/util/redux.js
ogupte: it should probably be organized such that all these util functions are at the top-level, where `dispatchTree` is passed as an arg to those which need to access it.

src/util/redux.spec.js
ogupte: ok cool

src/util/redux.spec.js
sodiumjoe: i thought we were on a version of mocha where you can return promises and it'll handle completion and errors correctly

ogupte: oh? maybe i didn't know that.

src/util/redux.js
sodiumjoe: ok, i can change that

src/util/redux.js
sodiumjoe: wut, it's right below

src/util/redux.spec.js
sodiumjoe: happy to, but it doesn't look like any of the other utils test files do this

ogupte: not yet...but they will...see the component-types test file

src/util/redux.js
sodiumjoe: in use, the dependency graph is going to look the same because the proposed functions have overlapping inputs (reducer tree). I think it's better to isolate the boundary between this utility and applications at a single point. I think we can adjust this after more real-world use, too.

src/util/redux.js
sodiumjoe: this is still hard because the functions that need access to `dispatchTree` are called before `dispatchTree` binding is assigned, so the reference doesn't propagate. It could be assigned to an empty "container" object and `dispatchTree` set to a property on that object, but that's pretty gross too.

src/util/redux.js
sodiumjoe: fixed

src/util/redux.spec.js
sodiumjoe: fixed

src/util/redux.spec.js
```diff
@@ -0,0 +1,208 @@
...
+					});
+				});
+
+				describe('thunks', () => {
+
+					beforeEach(() => dispatchTree.foo.asyncOperation('qux'));
+
+					it('should dispatch a thunk', () => {
+						const dispatchedThunk = mockDispatch.getCall(0).args[0];
+						assert(isFunction(dispatchedThunk))
```
ogupte: add a description here so it doesn't just say `AssertionError: true == false`

PR #256, opened 2016-05-18
Branding refresh and VerticalListMenu tweaks

### LINKED ISSUE #215: Improve VerticalListMenu to use divs instead of spans
- [x] `.lucid-VerticalListMenu-Item-content-text` should be a `div` instead of a `span`
- [x] we probably don't need [this][foo] anymore
- [x] make sure the new div is 100% width within the flex.

[foo]: https://github.com/appnexus/lucid/blob/v0.2.9/src/components/VerticalListMenu/VerticalListMenu.less#L41

### LINKED ISSUE #233: VerticalListMenu tweaks
- [x] With the work that's being done in #231, we should update `VerticalListMenu` to animate the left border on hover to match that style.
- [x] There are cases where it doesn't make sense to have items that are _expandable_ yet still style their hover as _selectable_. Come up with a good way to make that configurable.

jondlm:
- Fixed #233
- Fixed #215
- Updated `buildHybridComponent` to accept `reducers` as an option so I could override the behavior for the demo page

## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [x] IE11 (Win 7)
  - [x] Edge (Win 10)


src/docs/containers/color-palette.less
```diff
@@ -88,32 +89,30 @@
...
+		color: contrast(@@name);
+	}
 
 	// Recurse
 	.ColorPalette-loop(@i - 1);
 }
 
-.lucid-ColorPalette {
+.ColorPalette {
+	padding: 15px;
```
sodiumjoe: should this be a variable

src/docs/containers/color-palette.jsx
```diff
@@ -4,7 +4,7 @@ import React from 'react';
 
 import { lucidClassNames } from '../../util/style-helpers';
 
-const cx = lucidClassNames.bind('&-ColorPalette');
+const cx = lucidClassNames.bind('ColorPalette');
```
sodiumjoe: why not `lucid-` scope?

src/docs/index.jsx
```diff
@@ -225,11 +242,11 @@ const Component = React.createClass({
 			.value();
 
 		return (
-			<div>
+			<div className='Component'>
 				<h2>{componentName} {privateString} {composesComponents}</h2>
 				<div dangerouslySetInnerHTML={descriptionAsHTML} />
```
sodiumjoe: ooh dangerous

src/docs/index.jsx
```diff
@@ -352,36 +369,54 @@ const Component = React.createClass({
...
-				{_.map(items, (componentName) => {
+			<VerticalListMenu selectedIndices={[]}>
+				{_.map(_.sortBy(items), (componentName) => {
 					return (
-						<li key={componentName}>
-							<Link to={`/components/${componentName}`}>{componentName}</Link>
-						</li>
+						<Item
+							key={componentName}
+							onSelect={_.partial(this.goToPath, `/components/${componentName}`)}
```
sodiumjoe: why not just make these `<Router.Link>`s?

src/docs/containers/color-palette.jsx
jondlm: Cause our demo page is equivalent to a consuming application

src/docs/index.jsx
jondlm: Hah, not really

src/util/state-management.js
```diff
@@ -89,18 +89,18 @@ export function safeMerge (objValue, srcValue) {
...
 }
 
-export function buildHybridComponent(baseComponent, opts = {
-	setStateWithNewProps: true, // if true, new props will update state, else prop has priority over existing state
-	replaceEvents: false // if true, function props replace the existing reducers, else they are invoked *after* state reducer returns
-}) {
+export function buildHybridComponent(baseComponent, {
+	setStateWithNewProps = true, // if true, new props will update state, else prop has priority over existing state
+	replaceEvents = false, // if true, function props replace the existing reducers, else they are invoked *after* state reducer returns
+	reducers = _.get(baseComponent, 'definition.statics.reducers', {}),
```
ogupte: nice

src/docs/index.jsx
jondlm: Cause `Link`s render `<a>` tags and I want to navigate when the user selects an item in the `VerticalListMenu`

src/util/state-management.js
jondlm: Yup! I wrote a test for it too.

src/docs/containers/color-palette.less
jondlm: Yeah probably, I use it in lots of places. I've been cutting some corners on the demo page :/

src/docs/index.jsx
ogupte: i think it's bc VerticalListMenu.Item is the thing that should trigger navigation, not just link content of that item.

sodiumjoe: seems like this is a really common use-case that should interoperate nicely with `Router.Link`

PR #334, opened 2016-06-11
Default chart margins improvement

### LINKED ISSUE #325: Use default margins for charts
Currently charts accept a `margin` prop as an object. Usually consumers just want to override a single margin, but they unexpectedly have all their margins removed. We should merge their options with the default set so they don't have to explicitly pass every margin in.

jondlm:
Fixes #325 

`minor`: improved `LineChart` and `BarChart` to merge default margins with the `margin` prop so you can still retain defaults if you only want to override some of them

## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [x] IE11 (Win 7)
  - [x] Edge (Win 10)

### DIFF
--- a/src/components/BarChart/BarChart.jsx
+++ b/src/components/BarChart/BarChart.jsx
@@ -16,9 +16,6 @@ import Legend from '../Legend/Legend';
 
 const cx = lucidClassNames.bind('&-BarChart');
 
-const PADDING = 0.05;
-const PADDING_GROUPED_OR_STACKED = 0.3;
-
 const {
 	arrayOf,
 	func,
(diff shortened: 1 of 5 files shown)


src/components/BarChart/BarChart.jsx
```diff
@@ -163,11 +160,22 @@ const BarChart = createClass({
...
+			left: 80,
+		},
+	},
+
 	getDefaultProps() {
 		return {
 			height: 400,
 			width: 1000,
-			margin: {
+			margin: { // duplicated because `statics` aren't available during getDefaultProps
```
sodiumjoe: maybe better to leave the module scope `const`s and reference them here and in statics? not sure

jondlm: I tried that, but it makes it unclear what the defaults are when looking at the docs. It also isn't clear that they are available as statics. Tradeoffs :/ We should probably land on a pattern for this across our components.

PR #348, opened 2016-06-14
#347: Fixed Paginator bug that was causing textfield to revert to the default value onChange

### LINKED ISSUE #347: Paginator textfield reverts back to default value on update
**Reproduction Steps**
1. Click into Paginator textfield
2. Update page to `12`
3. Click away or hit enter

**Expected:** Textfield retains number entered by user.
**Actual:** Textfield reverts back to default value of `1`.

screen shot 2016-06-14 at 3 02 54 pm
screen shot 2016-06-14 at 3 03 05 pm
screen shot 2016-06-14 at 3 03 13 pm

reddyesha:
fixes #347 

## Notes on fix
`Paginator.handleTextFieldChange` expects `(value, totalPages)`. But instead, is receiving the following from `TextField.onSubmit`: `(value, {event, props: this.props })`.

In `Paginator.reducers.onPageSelect`, `selectedPageIndex` always gets set to `0` ->
`selectedPageIndex: _.clamp(pageIndex, 0, totalPages - 1)`, where `totalPages` is an object, not a number.

The fix was to calculate the correct value for `totalPages` in `Paginator.handleTextFieldChange` based on `totalCount` and `selectedPageSizeIndex` and pass that to `Paginator.reducers.onPageSelect`.

Not 100% sure about this. Do we need to pass the event?

## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [ ] IE11 (Win 7)
  - [ ] Edge (Win 10)

### DIFF
--- a/src/components/Paginator/Paginator.jsx
+++ b/src/components/Paginator/Paginator.jsx
@@ -117,11 +117,8 @@ const Paginator = createClass({
 		const {
 			onPageSelect,
 			selectedPageIndex,
-			totalCount,
-			pageSizeOptions,
-			selectedPageSizeIndex,
+			totalPages,
 		} = this.props;
-		const totalPages = _.ceil(totalCount/pageSizeOptions[selectedPageSizeIndex]);
 		const parsedPageNum = _.parseInt(pageNum);
 		if (_.isNaN(parsedPageNum)) {
 			return onPageSelect(selectedPageIndex, totalPages);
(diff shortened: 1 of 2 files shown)


src/components/Paginator/Paginator.jsx
```diff
@@ -113,10 +113,11 @@ const Paginator = createClass({
 		};
 	},
 
-	handleTextFieldChange(pageNum, totalPages) {
```
sodiumjoe: Hey, while you're here, I actually made an oversight in omitting the event from the event handler, can you pass that along to `onSelect` in both places as well?

```
handleTextFieldChange(pageNum, event) {
...
return onPageSelect(selectedPageIndex, totalPages, event);
```

reddyesha: Will do!

sodiumjoe: awesome thank you!

PR #367, opened 2016-06-22
feature/Sidebar

### LINKED ISSUE #235: Sidebar
Projects that would benefit:

* (ATG) Trader Home 3rd Party Data needs double grid with splitter between
* (PTG) PCR/Deals uses sidebars on most screens
* (PTG) PPR app may use sidebar for navigation on new AQ screen

* Styling
  * Styling of tab that shows when it is closed (icon flips correct direction)
  * Animation
    * When opening / closing
    * When user hovers on the tab (animation hints at functionality to the user)
    * See redlines for animation details
* Cursor change on hover
* Close on double click
* Resizable
  * Resizability can be disabled (but Sidebar can still snap open/close)

Link to redlines:
http://maloneydesignstudio.com/pres/appnexus/310/#home

Link to UX Audit of the Splitter:
https://corpwiki.appnexus.com/x/-4CcBQ

ogupte:
## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [x] IE11 (Win 7)
  - [x] Edge (Win 10)

Fixes #235


src/components/SplitVertical/SplitVertical.jsx
```diff
@@ -0,0 +1,392 @@
...
+				 */
+				children: node,
+			},
+		}),
+	},
+
+	getDefaultProps() {
+		return {
+			isExpanded: true,
+			isAnimated: false,
```
ogupte: `isAnimated` is defaulted to false unlike the default behavior for Sidebar

src/components/SplitVertical/SplitVertical.jsx
```diff
@@ -0,0 +1,392 @@
...
+			collapseShift: 0,
+			onResize: _.noop,
+		};
+	},
+
+	getInitialState() {
+		return {
+			isAnimated: false,
```
ogupte: having `isAnimated` in state is to make sure it doesn't animation on first render, so it's always `false` on first render.

src/components/SplitVertical/SplitVertical.jsx
```diff
@@ -0,0 +1,392 @@
...
+			leftPane: leftPaneRef,
+			rightPane: rightPaneRef,
+		} = this.refs;
+
+		const leftPaneElement = _.get(filterTypes(children, SplitVertical.LeftPane), 0, <SplitVertical.LeftPane />)
+		const rightPaneElement = _.get(filterTypes(children, SplitVertical.RightPane), 0, <SplitVertical.RightPane />)
+		let primaryElement, primaryRef;
+		let secondaryElement, secondaryRef;
+
+		if (leftPaneElement.props.isPrimary && !rightPaneElement.props.isPrimary) {
```
ogupte: this looks redundant but it's really just making sure that if the user sets `isPrimary` on both panes (a mistake), it will just do the default behavior, which is to set the right pane as the primary.

src/components/SplitVertical/SplitVertical.jsx
```diff
@@ -0,0 +1,392 @@
...
+			primary: primaryElement.props,
+			primaryRef,
+			secondary: secondaryElement.props,
+			secondaryRef,
+		};
+	},
+
+	applyDeltaToSecondaryWidth(dX, isExpanded, secondaryStartRect, secondaryRef, secondary, right, innerRef, primaryRef, collapseShift=0) {
+		if (isExpanded) {
+			secondaryRef.style.flexBasis = `${secondaryStartRect.width + dX * (secondary === right ? -1 : 1)}px`;
```
ogupte: with so many dom elements and logic paths in this function, it was easier to have it mutate the dom nodes directly rather than doing the mutation externally. (have to mutate the dom nodes here for perf reasons to shortcut the state -> rerender cycle)

src/components/SplitVertical/SplitVertical.jsx
```diff
@@ -0,0 +1,392 @@
...
+					'&-is-animated': isAnimated,
+				}, className)}
+			>
+				<div className={cx('&-inner')} ref='inner'>
+					<div
+						{...omitProps(leftPaneProps, SplitVertical.LeftPane)}
+						className={cx('&-LeftPane', {
+							'&-is-secondary': leftPaneProps === secondary,
+						}, leftPaneProps.className)}
+						style={_.assign({}, {
```
ogupte: not sure why i'm using `_.assign` here rather than the spread operator, but i'll try to switch it up

src/components/SplitVertical/SplitVertical.jsx
sodiumjoe: still seems redundant, could just reverse the conditional
```
if (rightElement.props.isPrimary) { ... } else { ... }
```

might be good to have some custom propType checking?

src/components/Icon/GripperVerticalIcon/GripperVerticalIcon.jsx
```diff
@@ -0,0 +1,50 @@
...
+/**
+ *
+ * {"categories": ["visual design", "icons"], "extend": "Icon", "madeFrom": ["Icon"]}
+ *
+ * A vertical gripper icon.
+ */
+const GripperVerticalIcon = createClass({
+	displayName: 'GripperVerticalIcon',
+
+	_lucidIsPrivate: true,
```
ogupte: private since it's only use is in the Sidebar component

src/components/Icon/Icon.jsx
```diff
@@ -88,15 +88,15 @@ const Icon = createClass({
 
 		return (
 			<svg
+				width={size}
+				height={size}
+				viewBox={viewBox}
+				preserveAspectRatio={aspectRatio}
 				{...passThroughs}
```
ogupte: reordered so that passThroughs are able to overwrite default props if needed

src/components/SplitVertical/examples/0.dev.jsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
```
ogupte: whoops i forgot to remove this file

src/components/Sidebar/Sidebar.jsx
```diff
@@ -0,0 +1,221 @@
...
+	number,
+	string,
+	oneOf,
+	oneOfType,
+} = React.PropTypes;
+
+/**
+ * {"categories": ["layout"]}
+ *
+ * `Sidebar` renders an collapsible, resizeable side bar panel next to primary content.
```
sodiumjoe: > an collapsible

src/components/SplitVertical/SplitVertical.jsx
ogupte: just reversing the conditional isn't equivalent, it affects what the default behavior is. Take a closer look. if If followed your suggestion, the default behavior would be invoked any time `rightPaneElement.props.isPrimary` is true, which is incorrect, because it needs to behave the same if both `rightPaneElement.props.isPrimary` and `leftPaneElement.props.isPrimary` are both ommitted.

sodiumjoe: seems better to capture the default value logic in `getDefaultProps` and assume they're both populated here

sodiumjoe: looks like you're already populating them in `getDefaultProps` actually, but if the default is for `right` to be primary, why not have it default to `true`

ogupte: the reason default props doesn't really help a whole lot in this case (it will be ignored by this logic) is because the props of those two different child components (LeftPane and RightPane) may be in conflict with each other, so it doesn't really matter what is returned by `getDefaultProps`, since it will be ignored.

src/components/Sidebar/Sidebar.jsx
```diff
@@ -0,0 +1,221 @@
...
+		/**
+		 * Render the Sidebar to the left or right of primary content.
+		 */
+		position: oneOf(['left', 'right']),
+		/**
+		 * Disable user resizing of the Sidebar.
+		 */
+		isResizeDisabled: bool,
+		/**
+		 * Set the title of the Sidebar. (alias for `Title` and `Sidebar.Title`)
```
jondlm: We haven't done this anywhere else, I'm not sure it's a good idea to introduce it into a single component.

src/components/Sidebar/Sidebar.less
```diff
@@ -0,0 +1,213 @@
+@size-Sidebar-icon: 20px;
```
jondlm: All of these should start with `@Sidebar-` to keep convention

jondlm: Also, you should use the `@size-expander-button: 20px;` variable for this one since it already exists.

src/components/SplitVertical/SplitVertical.jsx
jondlm: Might want to throw that in comment

src/components/SplitVertical/SplitVertical.jsx
jondlm: You should throw these comments in to code comments for posterity :)

src/components/SplitVertical/SplitVertical.less
```diff
@@ -0,0 +1,47 @@
+@SplitVertical-transition-duration: @timing-animation-fade;
+@SplitVertical-transition-half-duration: @SplitVertical-transition-duration / 2;
+
+.lucid-SplitVertical {
+	overflow: hidden;
+
+	& > &-inner {
```
jondlm: Is there a reason you're doing so many direct descendent selectors? Is it to support nested panes?

src/components/SplitVertical/SplitVertical.spec.jsx
```diff
@@ -0,0 +1,294 @@
...
+					done();
+				}, 10)
+			});
+
+			it('should apply the &-is-animated class when false', (done) => {
+				wrapper = mount(
+					<SplitVertical isAnimated={false} />
+				);
+
+				_.delay(() => {
```
jondlm: I feel like we should somehow tag these kinds of tests as potentially "flaky". I wrote a few of these for `TextInput` as well and they've failed on rare occasions due to timing issues.

src/components/SplitVertical/SplitVertical.less
ogupte: yes, nested frames break without the direct descendent selectors, and i think it's well formed css since it's styling a component with a very specific DOM structure.

src/components/SplitVertical/SplitVertical.spec.jsx
ogupte: i completely agree

src/components/Sidebar/Sidebar.jsx
ogupte: we haven't done it anywhere else because we never supported alias propNames until recently. I kept the upperCase propName to be in line with our other components, but i actually wanted to encourage consumers to use the prop in this case, which is why i also defined the lowercase propName.

PR #368, opened 2016-06-23
default Button type='button' - to avoid accidental trigger onEnter elsewhere

tarun:
## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [x] IE11 (Win 7)
  - [x] Edge (Win 10)

…ing Enter key on the page

### DIFF
--- a/src/components/Button/Button.jsx
+++ b/src/components/Button/Button.jsx
@@ -71,6 +71,10 @@ const Button = createClass({
 		 * Signature: `({ event, props }) => {}`
 		 */
 		onClick: func,
+		/**
+		 * form element type variations of button. Defaults to 'button' to avoid being triggered by 'Enter' anywhere on the page. Passed through to DOM button.
+		 */
+		type: string,
 	},
 
 	getDefaultProps() {
@@ -78,6 +82,7 @@ const Button = createClass({
 			isDisabled: false,
 			isActive: false,
 			onClick: _.noop,
+			type: 'button',
 		};
 	},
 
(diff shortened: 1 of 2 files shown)


tarun:
The issue - when hitting the enter key anywhere else on the page - text field, etc.
The onClick handler of the button is getting triggered... Ideally we want enter key to either submit the form or do nothing..
However Browsers default all buttons to type=submit by default - which wreaks havoc as hitting enter can trigger the onClick of any other button on the page  (technically the first in a form).

The generic use case of button is for form controls to modify UX behavior and other controls.. therefore this often leads to unexpected behavior.

By setting the default props as `<Button type=button />` - we avoid the accidental behavior in the 98% use case - and if a user wants to use a button to submit a form - they should explicitly specify `type=submit`
And this way we don't have to preventDefault on other form controls (text input) onEnter to avoid unexpected behavior.

src/components/Button/Button.jsx
```diff
@@ -71,13 +71,18 @@ const Button = createClass({
 		 * Signature: `({ event, props }) => {}`
 		 */
 		onClick: func,
+		/**
+		 * form element type variations of button. Defaults to 'button' to avoid being triggered by 'Enter' anywhere on the page. Passed through to DOM button.
+		 */
+		type: string,
```
sodiumjoe: is this not redundant with `passThroughs`? i.e. doesn't this already work?

jondlm: It is redundant, but I think this makes it clear that we are defaulting it to `button` aye?

sodiumjoe: good pt

jondlm: Good question, not sure about that

PR #370, opened 2016-06-23
276 loading indicator

### LINKED ISSUE #276: Loading Indicators
AC:
- A shape containing an icon (regular or animated SVG) of the developer/designer's choosing (ie could be loading indicator svg, could be success icon, etc). 
- ~~Appropriate components (*Chart, *Table, anything else?) have `loading` states available which leverage LoadingIndicator.~~  moved to #371 

Options: 
1. Title (note: in a real-worldconsole usage, we should always try to have a title)
2. Paragraph Text
3. Background cover – If this parameter is on (as it should be in most Console use cases), the background of whatever parent component this component exists within should be covered with a transparent black overlay. This helps the user understand the section of the page that this indicator relates to. (In some cases this might be the whole page).
~~4. Is Dismissable
--- Open Question: should we include a "Fade Away" parameter like we have on Banners? I think the most common use case for this is for a loading indicator, so I think most times it would be preferable that this whole component removes itself from the screen once no longer needed. So rather than a timed fade away, like on banners, this would need to accept some kind of trigger signal.~~

sodiumjoe:
fixes #276 

ACs for 276 include adding loading indicator to Charts and Tables, but that would be a breaking change, so I think it's ok for now for consumers to do:

```
<LoadingIndicator isLoading>
  <BarChart>...</BarChart>
</LoadingIndicator>
```
I've captured the other work #371 

## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [x] IE11 (Win 7)
  - [x] Edge (Win 10)

http://docspot.devnxs.net/projects/lucid/276-loading-indicator/#/components/LoadingIndicator


src/components/LoadingIndicator/LoadingIndicator.jsx
```diff
@@ -0,0 +1,93 @@
...
+
+		const { LoadingMessage } = LoadingIndicator;
+
+		const messageElement = _.first(findTypes(props, LoadingMessage)) || <LoadingMessage />;
+		const otherChildren = rejectTypes(children, [LoadingMessage]);
+
+		return (
+			<div
+				{...passThroughs}
+				className={cx('&', '&-container', className)}
```
ogupte: is the `&-container` class redundant?

src/components/LoadingIndicator/LoadingIndicator.jsx
```diff
@@ -0,0 +1,93 @@
...
+				className,
+				children,
+				isLoading,
+				...passThroughs,
+			},
+		} = this;
+
+		const { LoadingMessage } = LoadingIndicator;
+
+		const messageElement = _.first(findTypes(props, LoadingMessage)) || <LoadingMessage />;
```
ogupte: love it!

src/components/LoadingMessage/LoadingMessage.jsx
```diff
@@ -0,0 +1,87 @@
...
+
+/**
+ *
+ * {"categories": ["communication"], "madeFrom": ["LoadingIcon"]}
+ *
+ * A loading message.
+ *
+ */
+const LoadingMessage = createClass({
+	propName: 'LoadingMessage',
```
ogupte: i think it's safe to remove the propName here, the only valid value you can pass in as a prop is if it's `{<Element>}`, and at that point, might as well just pass it thru children.

src/components/LoadingMessage/LoadingMessage.jsx
```diff
@@ -0,0 +1,87 @@
...
+		const iconElement = _.first(findTypes(props, Icon));
+		const iconChildren = _.get(iconElement, 'props.children', <LoadingIcon />);
+		const titleElement = _.first(findTypes(props, Title));
+		const titleChildren = _.get(titleElement, 'props.children');
+		const bodyElement = _.first(findTypes(props, Body));
+		const bodyChildren = _.get(bodyElement, 'props.children', null);
+
+		return (
+			<div
+				{..._.omit(passThroughs, ['Icon', 'Title', 'Body', 'children'])}
```
ogupte: you can use `{ omitProps } from 'component-types'` for this purpose if you want

src/components/LoadingIndicator/LoadingIndicator.jsx
sodiumjoe: yep, i think you're right

src/components/LoadingMessage/LoadingMessage.jsx
sodiumjoe: good pt

src/components/LoadingMessage/LoadingMessage.jsx
sodiumjoe: oh nice!

src/components/LoadingMessage/LoadingMessage.less
```diff
@@ -0,0 +1,33 @@
+.lucid-LoadingMessage {
+
+	.lucid-Icon {
```
ogupte: why is this rule-set above all the declarations while the other nested rule-sets are below the declarations?

src/components/LoadingMessage/LoadingMessage.jsx
sodiumjoe: looks like i'll have to modify it to take multiple components

src/components/LoadingMessage/LoadingMessage.less
```diff
@@ -0,0 +1,33 @@
...
+		display: inline-block;
+	}
+
+	@LoadingMessage-padding: @size-standard;
```
ogupte: For other components we put the component-specific variables in the global scope rather than the component scope. do we want to break out of that pattern here?

src/components/LoadingMessage/LoadingMessage.less
sodiumjoe: no good reason

src/components/LoadingMessage/LoadingMessage.less
sodiumjoe: Hm, if that's the convention, i can stick to it, but i'm not sure why we wouldn't using real scopes when they're available.

src/components/LoadingMessage/LoadingMessage.less
```diff
@@ -0,0 +1,33 @@
...
+
+	line-height: 1em;
+	background-color: @color-white;
+	position: relative;
+	display: inline-block;
+	text-align: center;
+	border: @border-lightBorder;
+	border-radius: @size-borderRadius;
+	padding: @LoadingMessage-padding;
+	width: 135px + (@LoadingMessage-padding * 2);
```
ogupte: if you use `box-sizing: border-box`, you don't need to add the side paddings to the width here

src/components/LoadingMessage/LoadingMessage.spec.jsx
```diff
@@ -0,0 +1,122 @@
...
+import LoadingIcon from '../Icon/LoadingIcon/LoadingIcon';
+import SuccessIcon from '../Icon/SuccessIcon/SuccessIcon';
+
+const {
+	Icon,
+	Title,
+	Body,
+} = LoadingMessage;
+
+describe('LoadingMessage', () => {
```
ogupte: this isn't a big deal, but this test breaks the pattern of tests we have for most other components:

```
describe('component')
    describe('render')
    describe('props')
    describe('child components')
```

src/util/component-types.js
```diff
@@ -73,7 +73,7 @@ export function findTypes(props, types=[]) {
 
 	// get elements from props (using type.propName)
 	const elementsFromProps = _.reduce(types, (acc, type) => {
-		if (!_.isEmpty(type.propName)) {
+		if (!_.isUndefined(type.propName)) {
```
ogupte: thanks for this...this is subtle but important to be able to pass null as a value

src/components/LoadingMessage/LoadingMessage.less
sodiumjoe: oh nice

src/components/LoadingMessage/LoadingMessage.spec.jsx
sodiumjoe: will change

src/components/LoadingMessage/LoadingMessage.less
ogupte: i think we should talk more about whether or not to scope component-specific less variables...i'm not sure why we don't scope them either

sodiumjoe: `border-box` made this variable unnecessary, but we should still talk about this

PR #385, opened 2016-06-24
add SearchField

### LINKED ISSUE #259: SearchField
A common use case is a TextField with a build in Search icon and basic onClick handlers similar to Autocomplete for use in forms, and searching data tables.

sodiumjoe:
fixes #259 

http://docspot.devnxs.net/projects/lucid/259-search-field/#/components/SearchField

## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [x] IE11 (Win 7)
  - [x] Edge (Win 10)


src/components/SearchField/SearchField.jsx
```diff
@@ -0,0 +1,137 @@
...
+		const textFieldProps = {
+			isDisabled,
+			onChange,
+			onSubmit,
+			placeholder,
+			isSingleLine: true,
+			value,
+		};
+
+		const defaultIcon = <SearchIcon className={cx('&-Icon', { '&-Icon-active': !_.isEmpty(value) })} />;
```
sodiumjoe: possible optimization is to pull this up into a instance member, so it doesn't have create this element on every render

src/components/SearchField/SearchField.jsx
```diff
@@ -0,0 +1,137 @@
...
+				...passThroughs,
+			},
+		} = this;
+
+		const {
+			Icon,
+			TextField,
+		} = SearchField;
+
+		const textFieldProps = {
```
sodiumjoe: these are the props i thought would be most useful for the default case. if you need more than this, you can just pass in your own `SearchField.TextField`, which is just a raw `TextField` with whatever props you want

src/components/SearchField/SearchField.jsx
```diff
@@ -0,0 +1,137 @@
...
+		TextField,
+		Icon: createClass({
+			displayName: 'SearchField.Icon',
+			propName: 'Icon',
+		}),
+	},
+
+	reducers,
+
+	propTypes: {
```
sodiumjoe: should we have an `isLoading` prop that just changes the searchIcon to a loadingIcon?

src/components/SearchField/SearchField.jsx
```diff
@@ -0,0 +1,137 @@
...
+	node,
+	number,
+	oneOfType,
+	string,
+} = React.PropTypes;
+
+/**
+* {"categories": ["controls", "text"], "madeFrom": ["TextField", "SearchIcon"]}
+*
+* This is a container that renders panels and controls its expansion/retraction.
```
sodiumjoe: whoops

src/components/SearchField/SearchField.jsx
```diff
@@ -0,0 +1,138 @@
...
+			onChange,
+			onSubmit,
+			placeholder,
+			isSingleLine: true,
+			value,
+		};
+
+		const defaultIcon = <SearchIcon className={cx('&-Icon', { '&-Icon-active': !_.isEmpty(value) })} />;
+
+		const iconElement = _.first(findTypes(props, Icon));
```
sodiumjoe: we do `_.first(findTypes())` a lot, maybe we should have a `findFirstOfType`

src/components/SearchField/SearchField.jsx
ogupte: should wait until it's requested, right now there's a way a consumer can accomplish the same effect, if we notice a lot repeat usage like that, we can make it a new prop

src/components/SearchField/SearchField.jsx
ogupte: i think you're right!

src/components/SearchField/SearchField.jsx
ogupte: if you inlined it in the ternary, it would be short circuited. and never evaluate

src/components/SearchField/SearchField.less
```diff
@@ -0,0 +1,29 @@
...
+
+		&:focus + .lucid-SearchIcon {
+			fill: @color-primary;
+		}
+
+	}
+
+	&-Icon {
+		position: absolute;
+		top: @TextField-size-height / 2 - 8;
```
ogupte: instead of positioning like this, you should wrap the icon with something like

```
<div style="
	position: absolute;
	right: 0;
	top: 0;
	height: 100%;
	display: flex;
	align-items: center;
	margin-right: 6px;
"></div>
```

so that if the icon, or text field height changes it will always be centered vertically

src/components/SearchField/SearchField.jsx
sodiumjoe: still a lot of unnecessary renders in the default case?

src/components/ContextMenu/ContextMenu.jsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import _ from 'lodash';
 import Portal from '../Portal/Portal';
-import { createClass, findTypes } from '../../util/component-types';
+import { createClass, getFirst } from '../../util/component-types';
```
jondlm: Noice!

src/components/SearchField/SearchField.less
```diff
@@ -0,0 +1,29 @@
+.lucid-SearchField {
+
+	position: relative;
+
+	.lucid-TextField {
+
+		border-radius: @TextField-size-height / 2;
+		padding: 0 12px;
```
jondlm: Size variable here

halseyhemingway:
@sodiumjoe Search Icon should not turn teal unless there is text in the box. Border conveys that the user in "in" the field. Search Icon turning teal indicates that the search can take place (once there is some text in there)

sodiumjoe:
@halseyhemingway I fixed that, but it seems like that should maybe be configurable? I.e., some search endpoints require at least two characters, etc.

halseyhemingway:
Agree that making it configurable would be a nice option.

sodiumjoe:
@halseyhemingway done, docspot is updated with an example

PR #397, opened 2016-06-29
Fixes #374 - remove maxWidth on chart tooltips

### LINKED ISSUE #374: Remove max-width on chart tooltips
We have some use cases where series titles are very long. One solution is to remove the max-width restriction on tooltips inside charts.

image

jondlm:
`patch`: fixed an issue with tooltips and long field names on `BarChart` and `LineChart` by removing the max-width property on the underlying `ToolTip`

Fixes #374 

## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [x] IE11 (Win 7)
  - [x] Edge (Win 10)
- ~Unit tests written (`common` at minimum)~

### DIFF
--- a/src/components/Legend/Legend.less
+++ b/src/components/Legend/Legend.less
@@ -39,6 +39,7 @@
 	}
 
 	&-Item-indicator {
+		flex: 1 0 auto;
 		margin-right: @size-XS;
 	}
 }
--- a/src/components/LineChart/LineChart.jsx
+++ b/src/components/LineChart/LineChart.jsx
@@ -458,7 +458,12 @@ const LineChart = createClass({
 				{/* tooltips */}
 				<g transform={`translate(${margin.left}, ${margin.top})`}>
 					{hasToolTips && isHovering && !_.isNil(mouseX) ?
-						<ToolTip isExpanded={true}>
+						<ToolTip
+							isExpanded={true}
+							flyOutStyle={{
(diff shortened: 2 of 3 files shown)


src/components/Legend/Legend.less
```diff
@@ -39,6 +39,7 @@
 	}
 
 	&-Item-indicator {
+		flex: 1 0 auto;
```
jondlm: Fixed another small bug where the indicator could get squished with long titles.

src/components/LineChart/LineChart.jsx
```diff
@@ -458,7 +458,12 @@ const LineChart = createClass({
 				{/* tooltips */}
 				<g transform={`translate(${margin.left}, ${margin.top})`}>
 					{hasToolTips && isHovering && !_.isNil(mouseX) ?
-						<ToolTip isExpanded={true}>
+						<ToolTip
+							isExpanded={true}
+							flyOutStyle={{
```
sodiumjoe: there's actually a `flyOutMaxWidth` prop

sodiumjoe: looks like you'll have to change the proptype to `oneOf([number, string])`, tho

jondlm: Cool, I'll use that instead then.

PR #402, opened 2016-06-30
384 prop names prop types

### LINKED ISSUE #384: add common test to make sure `childComponent`s' `propName`s are included in `propTypes`

sodiumjoe:
fixes #384 

## PR Checklist

  - [x] Chrome
  - ~~[ ] Firefox~~
  - ~~[ ] Safari~~
  - ~~[ ] IE11 (Win 7)~~
  - ~~[ ] Edge (Win 10)~~

### DIFF
--- a/src/components/ButtonGroup/ButtonGroup.jsx
+++ b/src/components/ButtonGroup/ButtonGroup.jsx
@@ -27,7 +27,6 @@ const ButtonGroup = createClass({
 	components: {
 		Button: createClass({
 			displayName: 'ButtonGroup.Button',
-			propName: 'Button',
 		}),
 	},
 
--- a/src/components/ToolTip/ToolTip.jsx
+++ b/src/components/ToolTip/ToolTip.jsx
@@ -108,15 +108,12 @@ const ToolTip = createClass({
 	components: {
 		Target: createClass({
 			displayName: 'ToolTip.Target',
-			propName: 'Target',
 		}),
 		Title: createClass({
 			displayName: 'ToolTip.Title',
-			propName: 'Title',
 		}),
 		Body: createClass({
 			displayName: 'ToolTip.Body',
-			propName: 'Body',
 		}),
 	},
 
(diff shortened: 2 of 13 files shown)


sodiumjoe:
I made a lot of judgement calls on which child components should have `propName`s. The ones i took out were not documented, and didn't have any examples associated, so I'm calling this a `semver-none`.

src/components/ButtonGroup/ButtonGroup.jsx
```diff
@@ -27,7 +27,6 @@ const ButtonGroup = createClass({
 	components: {
 		Button: createClass({
 			displayName: 'ButtonGroup.Button',
-			propName: 'Button',
```
jondlm: What is the implication of removing a `propName`? Does that mean the child component shouldn't be pass in as a prop, but only as a child?

sodiumjoe: correct

src/components/ToolTip/ToolTip.jsx
```diff
@@ -108,15 +108,12 @@ const ToolTip = createClass({
 	components: {
 		Target: createClass({
 			displayName: 'ToolTip.Target',
-			propName: 'Target',
 		}),
 		Title: createClass({
 			displayName: 'ToolTip.Title',
-			propName: 'Title',
```
jondlm: Doesn't this make sense as a prop?

src/components/ToolTip/ToolTip.jsx
```diff
@@ -108,15 +108,12 @@ const ToolTip = createClass({
...
 		}),
 		Body: createClass({
 			displayName: 'ToolTip.Body',
-			propName: 'Body',
```
jondlm: Same here, what if you just want text?

sodiumjoe: The weird thing with these is that these wouldn't be props of `ToolTip`, they'd be props of a `ToolTip.FlyOut`, which would actually be the child component if this were factored correctly, but that depends on a refactor of `ContextMenu`.

jondlm: Gotcha, sounds good

PR #410, opened 2016-07-07
add child component documentation to docs page

### LINKED ISSUE #155: Documentation support for Child Components
Documentation generator/parser should support docs for Child Components that are defined along-side a main component.

sodiumjoe:
fixes #155 

Still need to add descriptions to child component defs.

## PR Checklist

http://docspot.devnxs.net/projects/lucid/155-docs-child-components/#/

  - ~~[ ] Chrome~~
  - ~~[ ] Firefox~~
  - ~~[ ] Safari~~
  - ~~[ ] IE11 (Win 7)~~
  - ~~[ ] Edge (Win 10)~~


gulp/docs.js
```diff
@@ -74,74 +69,185 @@ module.exports = {
...
+								documentation.set('description', description);
+							},
+							reactDocgen.handlers.displayNameHandler,
+							reactDocgen.handlers.componentMethodsHandler,
+							reactDocgen.handlers.componentMethodsJsDocHandler,
+						], function(handler) {
+							handler(documentation, definition);
+						});
+
+						documentation.set(
```
ogupte: not sure if we want to use the `displayName` or if it more correct to use the `${moduleName}.${childComponentPropertyName}

sodiumjoe: so this is a little weird because `displayName` here just means what we use to display in the docs, which i don't think is the same thing as the component `displayName` to be shown in react error stacks, etc. I was a little confused at first, too, and here I'm just sticking to what was already being done. I'm not sure what the right answer is.

gulp/docs.js
```diff
@@ -74,74 +69,185 @@ module.exports = {
...
-								});
+							visitImportDefaultSpecifier: function(path) {
+								importLocalNames.push(path.value.local.name);
 								return false;
 							},
-							visitExportDefaultDeclaration: function (path) {
-								exportIdentiferName = path.value.declaration.name;
+						});
+
+						recast.visit(ast, {
```
ogupte: can't you combine this `.visit` call with the last `.visit` call?

gulp/docs.js
```diff
@@ -74,74 +69,185 @@ module.exports = {
...
+							visitExportDefaultDeclaration: function(path) {
+								exportName = path.value.declaration.name;
+								recast.visit(ast, {
```
ogupte: not sure this nested `.visit` is necessary

gulp/docs.js
```diff
@@ -74,74 +69,185 @@ module.exports = {
...
+							_.forEach(reactDocgen.defaultHandlers, function(handler) {
+								handler(documentation, definition);
+							});
+
+							return recast.visit(definition, {
+								visitObjectExpression: function (path) {
+									var isPrivateComponent = _.chain(path.get('properties').value)
+										.find(_.matchesProperty('key.name', '_lucidIsPrivate'))
+										.get('value.value', false)
+										.value();
```
ogupte: no feedback, this is just funny: `value.value.value()`

gulp/docs.js
sodiumjoe: nah, because the order matters, i.e. i need `importLocalNames` to be fully populated before I can check if a refernce is to an imported component on L118

gulp/docs.js
ogupte: fair enough

gulp/docs.js
sodiumjoe: you're right, this could just be after

sodiumjoe: fixed

PR #414, opened 2016-07-07
use omitProps everywhere with passThroughs

### LINKED ISSUE #408: Fix pass throughs on a couple components
We need to omit props on a couple components. There are probably more than two, but theses are the ones we know about right now.

appnexus_console_-_console_report_ui

sodiumjoe:
kind of a doozy: http://docspot.devnxs.net/projects/lucid/408-passthroughs/#/

smoke-testing help appreciated

Fixes #408

## PR Checklist

  - [x] Chrome
  - ~~[ ] Firefox~~
  - ~~[ ] Safari~~
  - ~~[ ] IE11 (Win 7)~~
  - ~~[ ] Edge (Win 10)~~


src/components/Button/Button.jsx
```diff
@@ -114,7 +114,7 @@ const Button = createClass({
 
 		return (
 			<button
-				{...passThroughs}
+				{...omitProps(passThroughs, Button, ['callbackId'])}
```
sodiumjoe: this pattern is a little problematic, sort of couples `Button` to `ButtonGroup`. we should try to think of another way to pass arbitrary data to callbacks

src/components/Icon/ArrowIcon/ArrowIcon.jsx
```diff
@@ -46,7 +47,8 @@ const ArrowIcon = createClass({
 
 		return (
 			<Icon
-				{...passThroughs}
+				{...omitProps(passThroughs, ArrowIcon)}
+				{..._.pick(passThroughs, _.keys(Icon.propTypes))}
```
sodiumjoe: we want to allow `props` for `Icon` to get through so `Icon` can handle them

src/components/Button/Button.jsx
jondlm: Yeah agreed. I'm not keen on the pattern either. I recently had a little discussion on twitter about this. One solution was to use a memoized version of `partial`. I'm not sure I completely understand the implications of it, but it's food for thought.

https://twitter.com/davidlymanning/status/715993185403740160

src/components/RadioGroup/RadioGroup.spec.jsx
```diff
@@ -48,14 +48,6 @@ describe('RadioGroup', () => {
 			});
 		});
 
-		describe('onSelect', () => {
```
sodiumjoe: this was not a good test

1. i don't think we should test that `defaultProps` works properly
2. this was testing the exact wrong thing, i.e. that `onSelect` was getting passed to the root `span`, which we shouldn't be doing, and which `omitProps` prevents

src/components/Button/Button.jsx
sodiumjoe: we should get @tarun in on this discussion, i think the's thought about this the most

src/components/ContextMenu/ContextMenu.jsx
```diff
@@ -188,14 +188,16 @@ const ContextMenu = createClass({
...
 
-		const alignmentOffset = (alignment === ContextMenu.CENTER)
-			? (getAlignmentOffset(_.includes([ContextMenu.UP, ContextMenu.DOWN], direction) ? flyOutWidth : flyOutHeight))
-			: 0;
-
 		if (!flyOutPortal) {
 			return {};
 		}
 
+		const alignmentOffset = !_.isUndefined(this.props.alignmentOffset)
```
jondlm: I'm guessing this was the bugfix you mentioned

sodiumjoe: yep, wasn't setting `alignmentOffset` at all

src/components/DataTable/examples/2.column-groups.jsx
```diff
@@ -80,7 +80,7 @@ export default React.createClass({
 		return(
 			<DataTable
 				data={data}
-				hasExtraWhitespace
```
jondlm: Good catch

src/components/Dialog/Dialog.spec.jsx
```diff
@@ -5,15 +5,24 @@ import describeWithDOM from '../../util/describe-with-dom';
 import assert from 'assert';
 
 import Dialog from './Dialog';
+import Overlay from '../Overlay/Overlay';
 
-describe('Dialog', () => {
+describe.only('Dialog', () => {
```
jondlm: @sodiumjoe make sure to remove this :)

src/components/Dialog/Dialog.spec.jsx
```diff
@@ -14,6 +15,14 @@ describe('Dialog', () => {
 		},
 	});
 
+	it('should pass `isModal` to underlying Overlay', () => {
```
jondlm: Thanks for adding a test!

src/components/RadioGroup/RadioGroup.spec.jsx
jondlm: Agreed. There are probably a smattering of these in our tests.

src/components/Tabs/Tabs.jsx
```diff
@@ -132,7 +132,7 @@ const Tabs = createClass({
 							>
 								{_.get(getFirst(tabChildProp, Tabs.Title), 'props.children', '')}
 								{isProgressive && index !== tabChildProps.length - 1 ?
-									<svg className={cx('&-Tab-arrow')} xmlns='http://www.w3.org/2000/svg' width='8px' height='28px' viewBox='0 0 8 28' >
```
jondlm: Man, you found a ton of goodies in this PR.

src/components/Button/Button.jsx
tarun: here omitProps is just a call so the whole cost of creating partial function scenario most likely doesn't apply here...
as for memoization - a reselect.createSelector wrapped function may optimize - but probably not worth it at a lower level...

PR #426, opened 2016-07-19
add Breadcrumb

### LINKED ISSUE #378: Breadcrumb
https://corpwiki.appnexus.com/display/ux/Headers+Standardization

breadcrumb

sodiumjoe:
fixes #378 

http://docspot.devnxs.net/projects/lucid/378-breadcrumb/#/components/Breadcrumb

## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [x] IE11 (Win 7)
  - [x] Edge (Win 10)


src/components/Breadcrumb/examples/1.basic.jsx
```diff
@@ -0,0 +1,37 @@
...
+						</Item>
+						<Item key='list'>
+							<a href='#'>List</a>
+						</Item>
+						<Item key='item'>
+							Item
+						</Item>
+					</Breadcrumb>
+				</section>
+				<section style={{marginTop: 10}}>
```
ogupte: shouldn't this other example be in a new example file?

src/components/Icon/SeparatorIcon/SeparatorIcon.jsx
```diff
@@ -0,0 +1,45 @@
...
+/**
+ *
+ * {"categories": ["visual design", "icons"], "extend": "Icon", "madeFrom": ["Icon"]}
+ *
+ * An Separator icon.
+ */
+const SeparatorIcon = createClass({
+	displayName: 'SeparatorIcon',
+	propTypes: Icon.propTypes,
+
```
ogupte: i would set `_lucidIsPrivate: true` here unless there's another more public use case for this icon

sodiumjoe: good call

src/components/Breadcrumb/examples/1.basic.jsx
sodiumjoe: sure

src/components/Breadcrumb/Breadcrumb.less
```diff
@@ -0,0 +1,43 @@
...
+		display: flex;
+		flex-direction: row;
+		flex-wrap: wrap;
+		align-items: center;
+		font-size: @size-font;
+		color: @color-textColor;
+
+		a {
+			text-decoration: none;
+			color: @color-primary;
```
ogupte: should be using `@color-linkColor` and `@color-linkColorHover` on hover unless u intentionally left out a hover state

sodiumjoe: not intentional, should i pull those into a mixin?

src/components/Breadcrumb/Breadcrumb.less
```diff
@@ -0,0 +1,43 @@
...
+
+	}
+
+	.lucid-Icon {
+		height: @size-font;
+		width: @size-font;
+		display: block;
+	}
+
+	.lucid-Icon&-SeparatorIcon {
```
ogupte: this should be in `SeparatorIcon.less`

sodiumjoe: these are specific concerns of the icon in the context of `Breadcrumb`

src/components/Breadcrumb/Breadcrumb.less
```diff
@@ -0,0 +1,43 @@
+.lucid-Breadcrumb {
+
+	@breadcrumb-margin: 7px;
```
ogupte: can this value be derived from anything in variables.less, like `@size-XS + 1` maybe?

src/components/Breadcrumb/Breadcrumb.less
```diff
@@ -0,0 +1,43 @@
+.lucid-Breadcrumb {
```
ogupte: i think it's better to define the style more strictly like:
```
.lucid-Breadcrumb {
	& > .ludic-Breadcrumb-List {
		& > .lucid-Breadcrumb-Item {
			& > a {
			}
			& > .lucid-Icon {
				&.lucid-SeparatorIcon {
				}
			}
		}
	}
}
```
it ensures there's no weird styles applied to content that the consumer might not be expecting, and imo it's easier to maintain since it mirrors the element structure.

src/components/Breadcrumb/Breadcrumb.jsx
```diff
@@ -0,0 +1,90 @@
...
+								key={key}
+								className={cx('&-Item', props.className)}
+							>
+								{props.children}
+								<SeparatorIcon className={cx('&-SeparatorIcon')} />
+							</li>
+						))}
+						<li
+							{...lastItem.props}
+							key={lastItem.key}
```
ogupte: this is actually pretty cool

src/components/Breadcrumb/Breadcrumb.jsx
```diff
@@ -0,0 +1,90 @@
...
+		const items = findTypes(this.props, Breadcrumb.Item);
+		const initialItems = _.initial(items);
+		const lastItem = _.last(items);
+
+		return (
+			<nav
+				{...omitProps(passThroughs, Breadcrumb)}
+				className={cx('&', className)}
+			>
+				{!_.isEmpty(items) && (
```
ogupte: should use a ternary here instead of the logical operators. when the list of items is empty this block will render `{true}`. using a ternary is more explicit about rendering `{null}`

sodiumjoe: damn why are you so much smarter than me

src/components/Breadcrumb/Breadcrumb.less
sodiumjoe: i always thought this was better, too, but i thought we were trying to avoid deep nesting

src/components/Breadcrumb/Breadcrumb.less
sodiumjoe: idunno, it's not clear to me how it relates to other values, it's just a value in the redline. personally, i think it would make more sense for it to just be `@size-XS`

src/components/Breadcrumb/Breadcrumb.less
ogupte: not sure why you would avoid it, the compiler just flattens it out anyways. besides we're shouldn't be optimizing for best compile time.

PR #446, opened 2016-08-04
exporting generic tests

alisonailea:
## PR Checklist


jondlm:
On second thought I don't think we should do this. It means that `sinon` and `enzyme` become part of the regular dependencies, which would be yukky (technical term).

PR #473, opened 2016-08-22
strip synthetic events from action creator args in dev mode

sodiumjoe:
## PR Checklist

  - ~~[ ] Chrome~~
  - ~~[ ] Firefox~~
  - ~~[ ] Safari~~
  - ~~[ ] IE11 (Win 7)~~
  - ~~[ ] Edge (Win 10)~~

Our pattern of passing the underlying event as the last argument of
event handlers was causing generated action creators to get called with
the event, which at some point gets traversed, and the property access
throws a warning synthetic event pooling warning.

The events are never actually used by any of the reducers, so the
warning is benign afaict, so I'm only stripping them in dev mode.

~~The code is a little uglier than it could be because we're not
completely consistent about where the event goes (sometimes the last arg
is an object `{ props, event }`, sometimes the event itself).~~ fixed all the consistency issues

### DIFF
--- a/src/util/redux.js
+++ b/src/util/redux.js
@@ -227,7 +227,7 @@ function mergeProps(state, dispatchTree, ownProps) {
 	return _.merge({}, state, dispatchTree, ownProps);
 }
 
-function cleanArgs(args) {
+export function cleanArgs(args) {
 	return _.chain(args).last().has('event').value()
 		? _.dropRight(args)
 		: args;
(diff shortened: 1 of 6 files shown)


src/util/redux.js
```diff
@@ -222,3 +226,20 @@ function bindActionCreatorTree(actionCreatorTree, dispatch, path = []) {
...
+
+function hasEvent(obj) {
+	return obj && obj.event
+		? isEvent(obj.event)
+		: isEvent(obj);
+}
+
+function isEvent(obj) {
+	const proto = Object.getPrototypeOf(Object.getPrototypeOf(obj));
+	return proto && proto.constructor && proto.constructor.name === 'SyntheticUIEvent';
```
jondlm: `_.get(proto, 'constructor.name')`?

jondlm:
lgtm

PR #495, opened 2016-09-01
Adds styles to fix scrolling behavior with submarine nested in sidebar

### LINKED ISSUE #486: Issue with nested Sidebar and Submarine
[This codepen](http://codepen.io/njlg/pen/xEKKpd?editors=0110) doesn't behave as expected on Safari or FF. It appears to be an issue with overflows, but I'm not 100% sure.

mute:
## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [x] IE11 (Win 7)
  - [x] Edge (Win 10)

Fixes #486

### DIFF
--- a/src/components/SplitHorizontal/SplitHorizontal.jsx
+++ b/src/components/SplitHorizontal/SplitHorizontal.jsx
@@ -21,7 +21,7 @@ const {
 /**
  * {"categories": ["helpers"], "madeFrom": ["DragCaptureZone"]}
  *
- * `SplitHorizontal` renders a vertical split.
+ * `SplitHorizontal` renders a horizontal split.
  */
 const SplitHorizontal = createClass({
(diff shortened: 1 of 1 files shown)


src/components/SplitHorizontal/SplitHorizontal.jsx
```diff
@@ -387,6 +387,7 @@ const SplitHorizontal = createClass({
 					'&-is-animated': isAnimated,
 				}, className)}
 				style={{
+					display: 'flex',
```
jondlm: LGTM. @ogupte does this line up with what you were thinking?

ogupte: yeah looks good to me, not sure how i forgot to give both panes `overflow: 'auto'`, so this is a good catch!

PR #499, opened 2016-09-01
Adds a new `overlayKind` prop to LoadingIndicator.

### LINKED ISSUE #496: Add a lighter variant to LoadingIndicator
The overlay of LoadingIndicator has proven to be too dark for many use cases. We need the option for a "Light" version where the overlay is white and set to 85% opacity.

mute:
## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [x] IE11 (Win 7)
  - [x] Edge (Win 10)

Fixes #496

### DIFF
--- a/src/components/LoadingIndicator/LoadingIndicator.jsx
+++ b/src/components/LoadingIndicator/LoadingIndicator.jsx
@@ -9,6 +9,7 @@ const cx = lucidClassNames.bind('&-LoadingIndicator');
 const {
 	bool,
 	node,
+	oneOf,
 	string,
 } = React.PropTypes;
 
@@ -39,6 +40,13 @@ const LoadingIndicator = createClass({
 		 * Controls the visibility of the `LoadingMessage` and overlay.
 		 */
 		isLoading: bool,
+		/**
+		 * Style variations for the overlay behind the loading indicator.
+		 */
+		overlayKind: oneOf([
+			'light',
+			'dark',
+		]),
 	},
 
 	components: { LoadingMessage },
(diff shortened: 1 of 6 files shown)


mute:
Should this be `semver-minor` instead?

src/components/LoadingIndicator/LoadingIndicator.jsx
```diff
@@ -79,7 +89,7 @@ const LoadingIndicator = createClass({
 					transitionLeaveTimeout={300}
 				>
 					{isLoading && (
-						<div className={cx('&-message-container', { '&-has-overlay': hasOverlay })}>
+						<div className={cx('&-message-container', { [`&-has-overlay-${overlayKind}`]: hasOverlay })}>
```
mute: This could be a breaking change if anyone is selecting the overlay `div` by just the `has-overlay` container for styling, but I assumed that probably wasn't the case.

jondlm: Yep, you're right. We try to be thoughtful about our classname changes, but I don't think anyone is relying on this one. I'd mark it as a `semver-minor`.

sodiumjoe: any reason not to just add this class to the existing class?

jondlm: Good point, it could be just another class `&-kind-light` or something

mute: Sounds good to me -- I'll switch it.

PR #503, opened 2016-09-02
Remove Map

### LINKED ISSUE #502: Remove usage of `Map`
Internally we have some issues with the way babel is transpiling (or not transpiling) the `map` keyword. Let's get rid of it in favor of a plain object with string keys.

jondlm:
Fixes #502 remove usage of `Map` in favor of a POJO

## PR Checklist


src/components/Table/Table.jsx
```diff
@@ -578,7 +578,7 @@ function mapToGrid(trList, cellType='td', mapFn=_.property('element')) {
 
 
 	// iterate over each row
-	for (let rowIndex=0; rowIndex<cellRowList.length; rowIndex++) {
+	for (let rowIndex=0; rowIndex < cellRowList.length; rowIndex++) {
```
jondlm: This was messing up my syntax highlighter.

src/components/Table/Table.jsx
```diff
@@ -691,11 +691,11 @@ function renderRowsWithIdentifiedEdges(trList, cellType) {
 			}
 		}
 
-		if (!firstSingleLookup.has(rowIndex)) {
-			firstSingleLookup.set(rowIndex, false);
+		if (!_.has(firstSingleLookup, rowIndex)) {
```
jondlm: I probably didn't need to use `get` and `set` here, but why not?

PR #509, opened 2016-09-08
Fixes rendering issue in Firefox / Safari with Submarine nested in a …

### LINKED ISSUE #486: Issue with nested Sidebar and Submarine

mute:
## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [x] IE11 (Win 7)
  - [x] Edge (Win 10)

…Sidebar
Fixes #486


mute:
I tested this really thoroughly in all browsers and it's looking good. The 10th example for Sidebar demonstrates this functionality now.

I'd love another set of eyes on the changes, though, just in case there might be some functionality that I'm unaware of that might be broken.

src/components/Sidebar/Sidebar.less
```diff
@@ -169,6 +169,12 @@
 			& > .lucid-Sidebar {
 				flex: 1;
 			}
+
+			// Addresses a Firefox specific rendering issue with a Submarine nested in a Sidebar.
+			& > .lucid-Submarine, x:-moz-any-link, x:default {
```
jondlm: I'm not familiar with the `x:` thing, what is that?

mute: That's a really good question... I got it from http://browserhacks.com/ as a method of selecting for Firefox specifically, and based on my testing, it's accomplishing that.

I'll do some reading to try and figure out what it actually means.

mute: Also, the reason I needed this to be a browser specific selector is because applying these styles breaks rendering for the `PrimaryPane` in Safari, unfortunately.

jondlm: Dang, this _was_ a tricky one. Good to know!

jondlm:
@mute could you run `gulp docs-upload` so I can poke at the example?

mute:
Sure thing, docs are here: http://docspot.devnxs.net/projects/lucid/486-nested-sidebar-and-submarine-2/

PR #518, opened 2016-09-16
Adding a new kind of Expander

njlg:
Mike Brand has been working on this new pattern for expander-like content that is on a white background already. 

At first I thought it would just be a couple minor styles to the existing `Expander` component, but I had to make a small change and I'm not sure if that means I should build a new component. 

Please let me know what you think about this.

## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [x] IE11 (Win 7)
  - [x] Edge (Win 10)


src/components/Expander/Expander.jsx
```diff
@@ -109,25 +119,28 @@ const Expander = createClass({
...
 				{...omitProps(passThroughs, Expander)}
 				className={cx('&', {
 					'&-is-expanded': isExpanded,
+					'&-with-emphasis': withEmphasis,
 				}, className)}
 				style={style}
 			>
 				<header className={cx('&-header')} onClick={this.handleToggle}>
-					<span className={cx('&-icon')}>
+					{!withEmphasis ? (<span className={cx('&-icon')}>
```
jondlm: Since this is used in two places, make a component in this same file like this:

```javascript
const ChevronButton = createClass({
	propTypes: {
		isExpanded: bool,
	},

	render() {
		return (
			<span className={cx('&-icon')}>
				<ChevronIcon
					direction={this.props.isExpanded ? 'up' : 'down'}
				/>
			</span>
		);
	},
});
```

and then use it in both places like so:

```javascript
					{!withEmphasis ? <ChevronButton isExpanded={isExpanded} /> : null}
```

src/components/Expander/Expander.less
```diff
@@ -67,4 +67,35 @@
 	&-content {
 		overflow: hidden;
 	}
+
+	// kind
+	&-with-emphasis {
+
+		.lucid-Expander-header {
+			background: @color-backgroundColor;
+			border: 1px solid @color-borderColorLight;
```
jondlm: Make this `border: @border-lightBorder;`

src/components/Expander/Expander.less
```diff
@@ -67,4 +67,35 @@
...
+			padding: 12px;
```
jondlm: Make this `padding: @size-standard;`

src/components/Expander/Expander.less
```diff
@@ -67,4 +67,35 @@
...
+			display: flex;
+
+			.lucid-Expander-text {
+				font-size: 15px;
```
jondlm: This is another font size that should be reflected in `variable.less` but currently the naming isn't very good for font sizes so go ahead and leave this. More of a note to myself.

src/components/Expander/Expander.less
```diff
@@ -67,4 +67,35 @@
...
+				flex-shrink: 0;
+			}
+
+			&:hover .lucid-Expander-text > span {
+				text-decoration: none;
+			}
+		}
+
+		.lucid-Collapsible-content {
+			padding: 10px;
```
jondlm: I think this should be `@size-standard` as well to match the styling of `ExpanderPanel` and to align with the header.

src/components/Expander/examples/5.default-emphasis.jsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
```
jondlm: You can get rid of these two stateless examples since we've stopped adding them for the most part.

jondlm:
Not sure why the build is failing

src/components/Expander/Expander.less
mikebrand: 👍

jondlm:
Looks like travis isn't decrypting the browserstack credentials we have stored [here](https://github.com/appnexus/lucid/blob/v1.11.1/.travis.yml#L9). I wonder if it's because it's coming from a personal fork. I'll keep digging.

src/components/Expander/Expander.jsx
njlg: Cool. I updated the request.

src/components/Expander/Expander.less
njlg: I updated the request.

src/components/Expander/Expander.less
njlg: 👍

src/components/Expander/Expander.less
njlg: I updated the request.

src/components/Expander/examples/5.default-emphasis.jsx
njlg: I removed those two examples...

jondlm:
I've made a change to our travis setup that _should_ fix this. I'm testing it now.

jondlm:
Ahh dang, I don't think the build is going to pass. Turns out travis intentionally denies secure tokens from being exposed to personal forks, which totally makes sense actually:

> Encrypted environment variables are not available to pull requests from forks due to the security risk of exposing such information to unknown code.

[travis docs](https://docs.travis-ci.com/user/environment-variables/#Defining-encrypted-variables-in-.travis.yml)

njlg:
Closing in favor of #521

PR #521, opened 2016-09-19
Expander emphasis

njlg:
Moving PR #518 over inside the appnexus group (for successful builds)

## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [x] IE11 (Win 7)
  - [x] Edge (Win 10)


src/components/Expander/Expander.jsx
```diff
@@ -15,9 +15,34 @@ const {
 	func,
 	node,
 	object,
+	oneOf,
 	string,
 } = React.PropTypes;
 
+export const ChevronButton = createClass({
```
ogupte: I'm not sure this Component is necessary, it seems to be just a wrapper for ChevronIcon.

src/components/Expander/Expander.jsx
```diff
@@ -138,6 +171,7 @@ const Expander = createClass({
 							<span key={this._labelKey}>{labelChildProp.children}</span>
 						: null}
 					</ReactCSSTransitionGroup>
+					{withEmphasis ? (<ChevronButton isExpanded={isExpanded} />) : null}
```
ogupte: if the only difference between normal and emphasis is the order of the elements, it can be done in the css with `flex-direction: reverse`

src/components/Expander/Expander.jsx
```diff
@@ -81,12 +106,21 @@ const Expander = createClass({
 		 * the expander icon.
 		 */
 		Label: any,
+
+		/**
+		 * Kind of expander this should be
```
ogupte: add to the description the difference between 'normal' and 'emphasis'

src/components/Expander/Expander.less
```diff
@@ -67,4 +67,35 @@
...
+	&-with-emphasis {
+
+		.lucid-Expander-header {
+			background: @color-backgroundColor;
+			border: @border-lightBorder;
+			padding: @size-standard;
+			display: flex;
+
+			.lucid-Expander-text {
+				font-size: 15px;
```
ogupte: use the `@size-L` less variable

src/components/Expander/Expander.less
```diff
@@ -67,4 +67,35 @@
 	&-content {
 		overflow: hidden;
 	}
+
+	// kind
+	&-with-emphasis {
```
ogupte: className should be `&-has-emphasis`

src/components/Expander/Expander.less
```diff
@@ -67,4 +67,35 @@
...
+				font-weight: bold;
+				flex: 1;
+			}
+
+			.lucid-Expander-icon {
+				margin: 0;
+				flex-shrink: 0;
+			}
+
+			&:hover .lucid-Expander-text > span {
```
ogupte: the `> span` isn't necessary in the selector

src/components/Expander/Expander.jsx
jondlm: That's not a bad idea, probably wouldn't need the chevron button in that case

src/components/Expander/Expander.less
jondlm: I asked him to leave this at 15px actually. It _should_ have a `@font-size-medium` or something to handle this one. We need to improve our font size variables.

PR #522, opened 2016-09-19
Verticaltabs

njlg:
## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [x] IE11 (Win 7)
  - [x] Edge (Win 10)


src/components/VerticalTabs/VerticalTabs.jsx
```diff
@@ -0,0 +1,138 @@
...
+				/**
+				 * Determines if the Tab is selected.
+				 */
+				isSelected: bool,
+			},
+		}),
+		/**
+		 * Titles can be provided as a child or prop to a Tab.
+		 */
+		Title: createClass({
```
sodiumjoe: shouldn't this be a child component of `VerticalTabs.Tab`?

src/components/VerticalTabs/VerticalTabs.jsx
```diff
@@ -0,0 +1,138 @@
...
+			onSelect,
+			selectedIndex,
+			style,
+			...passThroughs,
+		} = this.props;
+
+		// Grab props array from each Tab
+		const tabChildProps = _.map(findTypes(this.props, VerticalTabs.Tab), 'props');
+
+		const selectedIndexFromChildren = _.findLastIndex(tabChildProps, {
```
sodiumjoe: is this a pattern taken from somewhere else? imo, I think we should just stick to `selectedIndex`

src/components/VerticalTabs/VerticalTabs.jsx
```diff
@@ -0,0 +1,138 @@
...
+		});
+
+		const actualSelectedIndex = selectedIndexFromChildren !== -1
+			? selectedIndexFromChildren
+			: selectedIndex;
+
+		return (
+			<div
+				{...omitProps(passThroughs, VerticalTabs)}
+				style={style}
```
sodiumjoe: any reason not to just leave `style` as one of the `passThroughs`?

src/components/VerticalTabs/VerticalTabs.reducers.js
```diff
@@ -0,0 +1,12 @@
+export default {
+	onSelect(state = {}, selectedIndex, { props = {} } = {}) {
+		if (props.isDisabled) {
```
sodiumjoe: imo, this logic should go into the view

src/components/VerticalTabs/examples/complex-title-stateful-progression.jsx
```diff
@@ -0,0 +1,70 @@
...
+		</svg>
+	</span>
+);
+
+export default React.createClass({
+	render() {
+		return (
+			<div>
+				<VerticalTabs>
+					<VerticalTabs.Tab Title={titleOne}>One content</VerticalTabs.Tab>
```
sodiumjoe: i think in examples we should stick to using the `propName` only for simple strings, and use child component composition for rich content

njlg:
@sodiumjoe Most of this (code & examples) comes from the [Tabs](https://github.com/appnexus/lucid/blob/master/src/components/Tabs/Tabs.jsx) component. I want this component to have the same interface as that component so it will be (a) easy to understand given you understand how `Tabs` works and (b) will be trivial to switch between the two components without any loss of functionality.

Should we apply your feedback to `Tabs` & `VerticalTabs`?

sodiumjoe:
@njlg definitely agree that it should have the same interface, but I think we should emphasize best practices around strings for propNames and composition for rich components. So, yeah, imo we should switch both these examples and the `Tabs` examples.

As for the child component and selectedIndex stuff, I think you're right, we should just keep it the same.

src/components/VerticalTabs/VerticalTabs.jsx
jondlm: Do we support nesting like that now?

ogupte: yeah the nesting of types is incidental to the finding of a particular type in an element's children/props, so nesting is purely a cosmetic thing

src/components/VerticalTabs/VerticalTabs.jsx
```diff
@@ -0,0 +1,138 @@
...
+				className={cx('&', className)}
+			>
+				<VerticalListMenu
+					selectedIndices={[actualSelectedIndex]}
+					onSelect={onSelect}
+				>
+					{_.map(tabChildProps, (tabChildProp, index) => {
+						return (
```
ogupte: can be cleaned up to return the element instead of having the `return` statement.

src/components/VerticalTabs/VerticalTabs.jsx
```diff
@@ -0,0 +1,138 @@
...
+							<VerticalListMenu.Item
+								className={cx('&-Tab', {'&-Tab-is-active': actualSelectedIndex === index})}
```
ogupte:
```
<VerticalListMenu.Item
    className={cx('&-Tab'
```
the className should match the subcomponent name either both should be `Item` or both should be `Tab`.

src/components/VerticalTabs/VerticalTabs.jsx
```diff
@@ -0,0 +1,138 @@
...
+								key={index}
+							>
+								<span className={cx('&-Tab-content')}>{_.get(getFirst(tabChildProp, VerticalTabs.Title), 'props.children', '')}</span>
```
ogupte: instead of having the _.get default to `''` have it default to `null` or even remove the default entirely

src/components/VerticalTabs/VerticalTabs.jsx
```diff
@@ -0,0 +1,138 @@
...
+							</VerticalListMenu.Item>
+						);
+					})}
+				</VerticalListMenu>
+				<div className={cx('&-content')}>
+					{_.get(tabChildProps[actualSelectedIndex], 'children', '')}
```
ogupte: should be:
```
{_.get(tabChildProps, [actualSelectedIndex, 'children'])}
```

src/components/VerticalTabs/VerticalTabs.reducers.js
ogupte: Yeah this shouldn't be referencing the last arg at all as it is only used by the consumer.

src/components/VerticalTabs/VerticalTabs.jsx
njlg: Cool. I changed `VerticalTabs` & `Tabs` to just let that fall through.

src/components/VerticalTabs/VerticalTabs.reducers.js
njlg: Fixed both `VerticalTabs` & `Tabs`.

PR #526, opened 2016-09-21
Add browser build of lucid

### LINKED ISSUE #525: Add browser build of Lucid
It would be nice to include a fully bundled (sans React and ReactDOM) in our distribution so we could expose the library through https://unpkg.com/.

jondlm:
Fixes #525 by adding a bundled build of lucid to the `npm run preversion` command.

## PR Checklist

- ~~PR has one of the `semver-` labels~~

### DIFF
--- a/scripts/build-bundle.sh
+++ b/scripts/build-bundle.sh
@@ -1,0 +1,6 @@
+#!/usr/bin/env bash
+set -e
+
+node_modules/.bin/webpack --config webpack.config.bundle.js &
+node_modules/.bin/webpack --config webpack.config.bundle.js --minify &
+wait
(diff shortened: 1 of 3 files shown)


scripts/build-bundle.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+set -e
+
+node_modules/.bin/webpack --config webpack.config.bundle.js &
+node_modules/.bin/webpack --config webpack.config.bundle.js --minify &
+wait
```
jondlm: I only learned about it today. The `&` at the end of the two previous lines runs them as background jobs, i.e. in parallel. The `wait` command will end the script when both have finished.

PR #532, opened 2016-09-23
524: Add Number Formatters, 519: Add yValue to yAxisTooltipFormatter signature

### LINKED ISSUE #519: Improve yAxisTooltipFormatter
There are cases where it would be nice to have to unformatted value passed to the `yAxisTooltipFormatter`. Let's add another argument to the callback signature that contains the raw y value.

Current signature: 

```
(yField, yValueFormatted) => {}
```

Proposed signature:

```
(yField, yValueFormatted, yValue) => {}
```

This will need to be done in three components:

- src/components/BarChart/BarChart.jsx
- src/components/Bars/Bars.jsx
- src/components/LineChart/LineChart.jsx

### LINKED ISSUE #524: Add number formatter to constants
As per discussion in this ticket: https://jira.corp.appnexus.com/browse/YMP-1733

```
Example:
487 -> 487
4,872 -> 4.9K
4,872,156 -> 4.87M
4,002, 156 -> 4M
```

reddyesha:
## Details

WIP. Would like feedback on how to handle allowing users to pass in different formatters to format axes and tooltips differently.

screen shot 2016-09-22 at 10 39 51 am

## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [x] IE11 (Win 7)
  - [x] Edge (Win 10)

Fixes #519
Fixes #524


src/components/LineChart/LineChart.jsx
```diff
@@ -226,11 +226,16 @@ const LineChart = createClass({
...
+		 * y-value.
 		 *
-		 * Signature: `(yField, yValueFormatted) => {}`
+		 * Signature: `(yField, yValueFormatted, yValue) => {}`
 		 */
 		yAxisTooltipFormatter: func,
+		/**
+		 * An optional function used to format data in the tooltips.
+		 */
+		yAxisTooltipDataFormatter: func,
```
reddyesha: To allow for different formatting in tooltips. Open to ideas.

src/components/LineChart/LineChart.jsx
```diff
@@ -255,6 +260,10 @@ const LineChart = createClass({
 		 */
 		y2AxisFormatter: func,
 		/**
+		 * An optional function used to format data in the tooltips.
+		 */
+		y2AxisTooltipDataFormatter: func,
```
reddyesha: To allow for different formatting in tooltips. Open to ideas.

src/components/LineChart/LineChart.jsx
```diff
@@ -435,14 +446,18 @@ const LineChart = createClass({
...
 			: xScale.tickFormat();
-		const yFinalFormatter = yAxisFormatter
-			? yAxisFormatter
-			: yScale.tickFormat();
-		const y2FinalFormatter = y2AxisFormatter
-			? y2AxisFormatter
-			: y2Scale
-				? y2Scale.tickFormat()
-				: _.identity;
+		const yFinalFormatter = yAxisTooltipDataFormatter
```
reddyesha: `yFinalFormatter` is only used for formatting tooltips.

src/components/Bars/Bars.jsx
```diff
@@ -286,7 +286,11 @@ const Bars = createClass({
 												pointKind={1}
 												color={_.get(colorMap, field, palette[(fieldIndex + colorOffset ) % palette.length])}
 											>
-												{yTooltipFormatter(_.get(legend, field, field), yFormatter(data[seriesIndex][field]))}
+												{
+													yTooltipFormatter(_.get(legend, field, field),
+													yFormatter(data[seriesIndex][field])),
+													(data[seriesIndex][field])
```
jondlm: this doesn't look correct, parens aren't lining up

sodiumjoe:
proposal: the props should be `yAxisFormatter`, `xAxisFormatter`, `tooltipTitleFormatter` and `toolTipBodyFormatter`, the last two having a signature of `(xData, yData) => formattedString`

the default for `tooltipTitleFormatter` would then be `(x, y) => yAxisFormatter(y)`, and for `tooltipBodyFormatter` would be `(x, y) => xAxisFormatter(x)`

I think that would provide the most flexibility for consumers, but with sane defaults

this is probably a breaking change, tho, so maybe we should save it for 2.0 milestone.

src/components/LineChart/LineChart.jsx
```diff
@@ -371,12 +380,14 @@ const LineChart = createClass({
 			yAxisTitleColor,
 			yAxisMin,
 			yAxisTooltipFormatter,
+			yAxisTooltipDataFormatter,
```
sodiumjoe: i think you could get rid of some of the conditional ternary stuff below by making them defaults here:

```
yAxisFormatter: yScale.tickFormat(),
yAxisTooltipFormatter: yAxisFormatter,
yAxisTooltipDataFormatter: yAxisTooltipFormatter,
```

src/constants/formatters.js
```diff
@@ -0,0 +1,9 @@
+import * as d3Format from 'd3-format';
```
sodiumjoe: shouldn't this be in `src/utils`?

reddyesha: The thought was to make this a constant that can be used externally, much like `chartConstants`.

src/components/Bars/Bars.jsx
reddyesha: Fixed.

src/components/LineChart/LineChart.jsx
reddyesha: This would have to be done as part of the breaking change, right?

src/constants/formatters.js
jondlm: Yeah I told her to put them here, but you're probably right `src/utils` is a better place since they technically aren't constants.

src/components/LineChart/LineChart.jsx
sodiumjoe: no, same format as you have now, just cleaning up the conditional logic

src/constants/formatters.js
reddyesha: Is the consensus to move these to `src/utils`? If so, will they still be accessible?

src/components/LineChart/LineChart.jsx
reddyesha: Gotcha. Adding a default value of `yScale.tickFormat()` means we have to move the variable definition (for `yScale`) above the props. Cool with that?

reddyesha: And it doesn't seem like you can set `yAxisTooltipDataFormatter: yAxisTooltipFormatter,`

reddyesha: It doesn't seem like you can set the default value to another prop.

sodiumjoe: hm, that's weird, you should be able to as long as the one you're setting it to is above the one you're setting the default on

reddyesha: Actually, this isn't the place to set default values for props. Seems like setting `yAxisTooltipFormatter = yAxisFormatter` would just overwrite the value passed in for `yAxisTooltipFormatter`. Am I missing something?

reddyesha:
Tests are passing locally. Not sure why the build is failing.

src/constants/formatters.js
reddyesha: Moved to `src/util`

reddyesha:
Good catch, @jondlm. Fixed-

screen shot 2016-09-27 at 8 44 45 am

jondlm:
I added more tests around the formatter function and slightly re-worked the second formatter to round out decimals and behave as expected for large numbers.

Issue #279, opened 2016-05-25
Autocomplete better support for async suggestions


ogupte:

astropunk:
Here are a couple of issues/potential QoL improvements I've noticed when implementing lucid's autocomplete with async suggestions.

- Enter inside of textinput doesn't fire a 0 index selection when there is an exact match on a suggestion.   This functionality has to be embedded inside of `onChange`.

- Default logic for `DropMenu onCollapse` behaves oddly when swapping out suggestion lists.

- `onSelect` fires an `onChange` before the `onSelect`.  This potentially creates a useless api request.  There doesn't seem to be a way of overriding this functionality through props.

- Some way to tune debouncing `onChange`?

- No suggestions loading indicator

-  `Suggestions` prop is a flat array of strings.  I image most api backed autocompletes will translate the selection from the name string to an ID. This means that the app now has to track an array for the suggestion strings and an array of the api objects separately.  It could be nice if it also accepted an array of objects coupled  with a display key, or formatter function.
```
let mySuggestionsFromApi = [{id: 1, name: 'foo'}, {id:2, name: 'bar'}];
//...
suggestions={mySuggestionsFromApi}
suggestionFormatter={(suggestion) => {suggestion.name}}
onSelect={(index) => {
  goStoreTheIdOfMySelection(mySuggestionsFromApi[index].id);
}}
```

jacklenehan:
On hold until UX can resolve patterns around filtering/selecting/searching

PR #542, opened 2016-09-28
Adds a Searchable SingleSelect

### LINKED ISSUE #479: SingleSelect w/ search
A few things to note about this:

- I'm not sure if it should be an entirely new component or just an addition to `SingleSelect`
- We need to make sure it has an `isLoading` prop to handle asynchronous loading of options @dmaloney01 might need to weigh in on that unless he just wants the `LoadingIndicator` to be used

dropdown_singleselect_searchfilter_styling

mute:
## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [x] IE11 (Win 7)
  - [x] Edge (Win 10)

Fixes #479.

Docs here: http://docspot.devnxs.net/projects/lucid/479-single-select-w-search/#/components/SearchableSelect


mute:
There's still a lot of room for improvement to remove duplication between this and `SingleSelect`. I haven't started on that effort yet.

src/components/SearchableSelect/SearchableSelect.jsx
```diff
@@ -0,0 +1,427 @@
...
+							)}
+						>
+							{isItemSelected ? flattenedOptionsData[selectedIndex].optionProps.children : placeholder}
+						</span>
+						<CaretIcon direction={isExpanded ? direction : 'down'} />
+					</div>
+					{
+						isExpanded &&
+						<div className={cx('&-Search-container')} onClick={this.handleClickSearchField}>
+							<SearchField
```
mute: Maybe I should expose a prop that allows for passing through `SearchField` props so that users of the component can have more control over the SearchField itself?

src/components/SearchableSelect/SearchableSelect.jsx
```diff
@@ -0,0 +1,427 @@
...
+} = React.PropTypes;
+
+/**
+ *
+ * {"categories": ["controls", "selectors"], "madeFrom": ["DropMenu"]}
+ *
+ * A selector control (like native `<select>`) which is used to select a single option from a dropdown list.
+ * Supports option groups with and without labels.
+ */
+const SearchableSelect = createClass({
```
mute: I'm not tied to this name -- I just had to call it something so I could get this PR open.

sodiumjoe: I vote for `SingleSelectFilterable`. There were some questions about async search using this component, but I don't think that makes sense given that it doesn't do paging or infinite scroll or anything. I think this component is really only appropriate for cases where you can show the entire list at once, i.e. places where you could use an regular html `select`.

sodiumjoe:
> There's still a lot of room for improvement to remove duplication between this and SingleSelect. I haven't started on that effort yet.

Are you planning to merge first and factor out common stuff later?

mute:
That was my plan, since it seemed like we needed this component available pretty soon.

src/components/SearchableSelect/SearchableSelect.jsx
mute: I think part of the goal with this component is to address some of the concerns noted here: https://github.com/appnexus/lucid/issues/279 regarding async behavior.

I totally agree, though, that doing async loading of options with this component is risky territory.

When you say 'show the entire list at once', do you mean that you think that there shouldn't even be scrolling within the Flyout? Or just that the entire list of Options should be provided at mount and then never change?

src/components/SearchableSelect/SearchableSelect.jsx
```diff
@@ -0,0 +1,427 @@
...
+			isLoading: false,
+			optionFilter: this.defaultOptionFilter,
+			richChildRenderer: _.identity,
+			searchText: null,
+			selectedIndex: null,
+			DropMenu: DropMenu.getDefaultProps(),
+		};
+	},
+
+	getInitialState() {
```
sodiumjoe: talked offline about removing some of these optimizations

src/components/SearchableSelect/SearchableSelect.jsx
```diff
@@ -0,0 +1,427 @@
...
+		 * If `true`, option will be visible. If `false`, the option will not be visible.
+		 */
+		optionFilter: func,
+		/**
+		 * This function will be run against Options with Rich Content for children to allow for special 'underlining' behavior.
+		 * The default behavior of the function is to just return the rich content as is.
+		 *
+		 * Has the signature `(children, searchText)`
+		 */
+		richChildRenderer: func,
```
mute: That's true; I'll remove it.

The original intent of this function was to provide a function to use for managing the underlining of rich child contents. Plus, I just liked the name `richChildRenderer`.

src/components/DropMenu/DropMenu.jsx
```diff
@@ -465,26 +482,39 @@ const DropMenu = createClass({
...
+							!_.isEmpty(headerProps) &&
+							<div
+								{...headerProps}
+								className={cx('&-Header', headerProps.className)}
+								onKeyDown={this.handleKeydown}
+							/>
 						}
+						<div
+							className={cx('&-option-container')}
+							style={_.assign({}, flyOutStyle, optionContainerStyle)}
```
mute: I needed to do this to maintain compatibility with the existing functionality; otherwise, even if the flyOutStyle's maxHeight is set to a smaller value, the optionContainerStyle will ignore it.

mute:
After testing integrating this with the KeyValueTargeting module, I realized the way I'm firing the `onSelect` event needs a tweak to be consistent with how `SingleSelect` handles it, so hold off on merging this.

PR #545, opened 2016-09-30
Updates label alignment for long labels.

### LINKED ISSUE #529: Improve alignment of Labeled components for long inputs
Sometimes we have really long labels in small spaces. The current styling for labeled components doesn't handle this well. We need to adjust the alignment for the cases where the label is wrapping. This probably applies to the following components:

- `RadioButtonLabeled`
- `CheckboxLabeled`
- `SwitchLabeled` 

They should look like something like this:

image

mute:
## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [x] IE11 (Win 7)
  - [x] Edge (Win 10)

Fixes #529.
Docs here: http://docspot.devnxs.net/projects/lucid/529-improve-label-alignment/#/components/RadioButtonLabeled

This doesn't make a change to `SwitchLabeled` because it doesn't look like the text for that component wraps with longer labels.

### DIFF
--- a/src/components/CheckboxLabeled/CheckboxLabeled.jsx
+++ b/src/components/CheckboxLabeled/CheckboxLabeled.jsx
@@ -94,11 +94,10 @@ const CheckboxLabeled = createClass({
 						onSelect={onSelect}
 						{...omitProps(passThroughs, CheckboxLabeled)}
 				/>
-				{
-					labelChildProps
-							? labelChildProps.children || labelChildProps
-							: null
-				}
+				<div
+					{...labelChildProps}
+					className={cx('&-label')}
+				/>
 			</label>
 		);
 	},
--- a/src/components/CheckboxLabeled/CheckboxLabeled.less
+++ b/src/components/CheckboxLabeled/CheckboxLabeled.less
@@ -12,7 +12,7 @@
 	}
 
 	&-label {
-		align-self: center;
+		padding-top: @size-XXS;
 	}
 
 	&&-is-disabled {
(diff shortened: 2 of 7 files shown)


jondlm:
The styling doesn't quite look right. It's a little more tricky than it first looks, and frankly I'm not even sure exactly how we'll achieve it. Check out this picture to see what I mean:

image

The first line should be aligned the same regardless of wrapping.

mute:
Yeah, agreed -- I noticed the same thing. This was as close as I was able to get it without beginning to do some Javascript based detection of whether a line of text is wrapping or not.

I guess the question is whether this slightly off styling is an improvement over the existing wrapping styling.

jondlm:
That's a good question. @dmaloney01 it might be prohibitively difficult to get the exact alignment you were hoping for. Could you take a peek at the current fix and let us know if it's a worthwhile improvement?

dmaloney01:
@jondlm I think this is an improvement, but I do want to check on thing: Can I see a few long label examples in a row in the same grouping? One of the issues we were trying to address with this fix is a making sure there is a clear visual distinction between items in a Checkbox or Radio Button list.

mute:
@dmaloney01: I've updated the docs here: http://docspot.devnxs.net/projects/lucid/529-improve-label-alignment/#/components/CheckboxLabeled and here: http://docspot.devnxs.net/projects/lucid/529-improve-label-alignment/#/components/RadioButtonLabeled with some additional "long label" examples.

dmaloney01:
@mute Thanks! That helps. If it's not feasible to get the exact alignment we talked about, this is still a big improvement over what exists currently. This has my approval.

sodiumjoe:
I think you should be able to achieve this with a combination of `min-height` and `padding-top` on the label container div

jondlm:
Actually I think Joe's right. Removing `align-self: center` and adding `padding-top` looks correct to me.

mute:
This is updated now to use `padding-top`.

PR #571, opened 2016-10-12
#489: Create generic EmptyStateWrapper for when DataTable and Charts …

### LINKED ISSUE #489: Please add an empty state to the chart component
Let's create the empty state window for the charts. 

AC
* A window displays information about the no data state. This window lives above the chart layer. (the window is visible regardless of which tab is selected)
* striped 'no data' pattern appears at the top of the window
* window can contain a customizable Title
* window can contain customizable body text
* window can contain customizable image (maximum 375 x 200 px)
* window can contain customizable button (call to action)
* window width can be dependent on the width of the content, max width for window with text only is 375 px. 
* window should display in vertical and horizontal center of the chart area

designs with some specs - http://maloneydesignstudio.com/pres/appnexus/319
thank you
R

reddyesha:
## Changes

- Create generic `EmptyStateWrapper` for when `DataTable`/`LineChart`/`BarChart` doesn't have data.
- Add `isLoading` to `BarChart` and `LineChart`
- Renamed example files in `LineChart` and `BarChart` to be of the following format: `01.*.jsx`. This keeps examples in order. Can make this change for other component examples if needed.

## Chart changes
screen shot 2016-10-12 at 3 32 33 pm
screen shot 2016-10-12 at 3 32 45 pm
screen shot 2016-10-12 at 3 32 51 pm
screen shot 2016-10-12 at 3 32 57 pm

## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [x] IE11 (Win 7)
  - [x] Edge (Win 10)

Fixes #489


src/components/EmptyStateWrapper/EmptyStateWrapper.less
```diff
@@ -8,6 +8,7 @@
 
 	&-message-contents {
 		padding: @size-S @size-standard @size-standard @size-standard;
+		text-align: center;
```
reddyesha: See mock: http://maloneydesignstudio.com/pres/appnexus/319/#next

src/components/BarChart/BarChart.jsx
```diff
@@ -256,6 +274,23 @@ const BarChart = createClass({
 		};
 	},
 
+	components: {
+		/**
+		 * Body content for the message to display when the data table has no data.
+		 */
+		EmptyMessageBody: createClass({
+			displayName: 'DataTable.EmptyMessageBody',
```
mute: Should be `BarChart.EmptyMessageBody`

src/components/BarChart/BarChart.jsx
```diff
@@ -256,6 +274,23 @@ const BarChart = createClass({
...
+			propName: 'EmptyMessageBody',
+		}),
+		/**
+		 * Title text for the message to display when the data table has no data.
+		 */
+		EmptyMessageTitle: createClass({
+			displayName: 'DataTable.EmptyMessageTitle',
```
mute: Should be `BarChart.EmptyMessageTitle`

src/components/BarChart/BarChart.jsx
```diff
@@ -256,6 +274,23 @@ const BarChart = createClass({
 		};
 	},
 
+	components: {
+		/**
+		 * Body content for the message to display when the data table has no data.
```
mute: `...chart has no data.`

src/components/BarChart/BarChart.jsx
```diff
@@ -256,6 +274,23 @@ const BarChart = createClass({
...
+		 */
+		EmptyMessageBody: createClass({
+			displayName: 'DataTable.EmptyMessageBody',
+			propName: 'EmptyMessageBody',
+		}),
+		/**
+		 * Title text for the message to display when the data table has no data.
```
mute: `...chart has no data.`

src/components/BarChart/BarChart.jsx
```diff
@@ -335,6 +358,46 @@ const BarChart = createClass({
 			? yAxisTooltipDataFormatter
 			: yAxisFinalFormatter;
 
+		if (_.isEmpty(data) || width < 1 || height < 1) {
```
mute: I think you need to add `isLoading` to this condition also, so that it's possible to show loading even if the chart already has data.

src/components/EmptyStateWrapper/EmptyStateWrapper.jsx
```diff
@@ -83,8 +83,8 @@ const DataTableWrapper = createClass({
 			isLoading,
 		} = this.props;
 
-		const emptyMessageBodyProp = _.get(getFirst(this.props, DataTableWrapper.EmptyMessageBody), 'props');
-		const emptyMessageTitleProp = _.get(getFirst(this.props, DataTableWrapper.EmptyMessageTitle), 'props', {children: 'You have no Line Items.'});
+		const emptyMessageBodyProp = _.get(getFirst(this.props, EmptyStateWrapper.EmptyMessageBody), 'props');
+		const emptyMessageTitleProp = _.get(getFirst(this.props, EmptyStateWrapper.EmptyMessageTitle), 'props', {children: 'You have no Line Items.'});
```
mute: I think we should change the default here from 'You have no Line Items.' to 'You have no data.'.

src/components/LineChart/LineChart.jsx
```diff
@@ -361,6 +379,23 @@ const LineChart = createClass({
 		};
 	},
 
+	components: {
+		/**
+		 * Body content for the message to display when the data table has no data.
```
mute: Same thing with `data table` -> `LineChart` in this section.

src/components/BarChart/BarChart.jsx
```diff
@@ -256,6 +274,23 @@ const BarChart = createClass({
 		};
 	},
 
+	components: {
+		/**
+		 * Body content for the message to display when the chart has no data.
+		 */
+		EmptyMessageBody: createClass({
```
sodiumjoe: I think these could just be references to `EmptyStateWrapper.Body` and `EmptyStateWrapper.Title`

src/components/EmptyStateWrapper/EmptyStateWrapper.jsx
```diff
@@ -48,30 +48,30 @@ const DataTableWrapper = createClass({
...
 		 */
 		EmptyMessageTitle: any,
 	},
 
 	components: {
 		/**
-		 * Body content for the message to display when the data table has no data.
+		 * Body content for the message to display when there is no data.
 		 */
 		EmptyMessageBody: createClass({
```
sodiumjoe: I think this should just be `Body` and `Title`

src/components/BarChart/BarChart.jsx
```diff
@@ -335,6 +358,46 @@ const BarChart = createClass({
...
+
+			return (
+				<EmptyStateWrapper
+					isEmpty={_.isEmpty(data)}
+					isLoading={isLoading}
+				>
+					<EmptyStateWrapper.EmptyMessageBody {...emptyMessageBodyProp}/>
+					<EmptyStateWrapper.EmptyMessageTitle {...emptyMessageTitleProp}/>
+					<svg
+						{...omitProps(passThroughs, BarChart)}
```
sodiumjoe: hm, this seems correct, but it breaks our pattern of `passthroughs` going to the root element. Anyone else have an opinion?

mute: I feel like in this case and cases like this, even though the `Wrapper` is hierarchically the root element, my expectation as a user of the component would be that the root element is the `<svg>`. It feels like kind of an exception to the pattern.

src/components/LineChart/LineChart.jsx
```diff
@@ -496,14 +519,53 @@ const LineChart = createClass({
...
+			const emptyMessageBodyProp = _.get(getFirst(this.props, LineChart.EmptyMessageBody), 'props');
+			const emptyMessageTitleProp = _.get(getFirst(this.props, LineChart.EmptyMessageTitle), 'props', {children: 'You have no data.'});
+
+			return (
+				<EmptyStateWrapper
+					isEmpty={_.isEmpty(data)}
+					isLoading={isLoading}
+				>
+					<EmptyStateWrapper.EmptyMessageBody {...emptyMessageBodyProp} />
+					<EmptyStateWrapper.EmptyMessageTitle {...emptyMessageTitleProp} />
```
ogupte: I think these can be simplified so that the EmptyMessage children don't even need to be passed in if they don't exist:

```
const emptyMessageBody = getFirst(this.props, LineChart.EmptyMessageBody);
```

```
{EmptyMessageBody || null}
```

Currently they are always passed it with or without extra props, but they don't have to be.

src/components/BarChart/BarChart.jsx
```diff
@@ -335,6 +358,46 @@ const BarChart = createClass({
...
+		if (_.isEmpty(data) || width < 1 || height < 1 || isLoading) {
+			const emptyMessageBodyProp = _.get(getFirst(this.props, BarChart.EmptyMessageBody), 'props');
+			const emptyMessageTitleProp = _.get(getFirst(this.props, BarChart.EmptyMessageTitle), 'props', {children: 'You have no data.'});
+
+			return (
+				<EmptyStateWrapper
+					isEmpty={_.isEmpty(data)}
+					isLoading={isLoading}
+				>
+					<EmptyStateWrapper.EmptyMessageBody {...emptyMessageBodyProp}/>
```
ogupte: see comment in LineChart: https://github.com/appnexus/lucid/pull/571/files?diff=split#r83692560

src/components/BarChart/BarChart.jsx
ogupte: there are other components that have this conflict between root passThrus vs expectec passThrus, and so far, we kept them at the root level only, and created props that pass thru to the underlying elements as needed. we should keep with that pattern until we reach a major version upgrade.

mute: If we change it so that the passThroughs go to the `EmptyStateWrapper` now, won't that be a breaking change also though?

src/components/BarChart/BarChart.jsx
reddyesha: So, move the logic from lines 362 and 363 down here?

reddyesha: Nevermind, I get what you're saying.

src/components/BarChart/BarChart.jsx
reddyesha: Done

PR #590, opened 2016-10-20
fix className handling in CheckboxLabeled

sodiumjoe:
## PR Checklist


src/components/CheckboxLabeled/CheckboxLabeled.jsx
```diff
@@ -83,23 +83,23 @@ const CheckboxLabeled = createClass({
...
 			<label
-					className={cx('&', {
-						'&-is-disabled': isDisabled,
-						'&-is-selected': isIndeterminate || isSelected,
-					}, className)}
-					style={style}
+				className={cx('&', {
+					'&-is-disabled': isDisabled,
+					'&-is-selected': isIndeterminate || isSelected,
+				}, className)}
```
sodiumjoe: this is kind of weird, `this.props.className` gets passed to both the root `label` and the `Checkbox`, not sure how to handle this

sodiumjoe: most correct thing to do seems like to expose `Checkbox` as a child component

PR #593, opened 2016-10-21
allow validation styling without an error message

sodiumjoe:
use case here is sometimes I want to show an invalid state for `TextFieldValidated` for example, but without a message, because I have the validation message at a parent component level. So I want the styling, but I don't want the extra div because that messes with layout.

## PR Checklist

  - [x] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] IE11 (Win 7)
  - [ ] Edge (Win 10)
- [ ] ~~One core team UX approval~~

### DIFF
--- a/src/components/Validation/Validation.jsx
+++ b/src/components/Validation/Validation.jsx
@@ -35,6 +35,9 @@ const Validation = createClass({
 		 * In most cases this will be a string, but it also accepts any valid React
 		 * element. If this is a falsey value, then no error message will be
 		 * displayed.
+		 *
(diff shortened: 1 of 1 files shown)


src/components/Validation/Validation.jsx
```diff
@@ -67,7 +67,7 @@ const Validation = createClass({
 				}, className)}
 			>
 				{children}
-				{errorChildProps && errorChildProps.children ?
+				{errorChildProps && errorChildProps.children && errorChildProps.children !== true ?
```
jondlm: Might want to mention this new behavior in the comment for `Error`

Issue #614, opened 2016-11-03
Typing in SearchField causes uncontrolled text input warning


ogupte:
<img width="493" alt="screen shot 2016-11-03 at 2 16 30 pm" src="https://cloud.githubusercontent.com/assets/1967266/19985535/54c7a5a2-a1d0-11e6-8870-fccd364363ea.png">

jondlm:
Looks like we're still seeing this pop up in some cases with `TextField`.

PR #619, opened 2016-11-08
Converts to using Jest from Karma.

### LINKED ISSUE #612: Switch to `jest` as a test runner
Let's switch to using `jest` for our test runner.

It will give us:
- Snapshot Testing
- Coverage Information
- More sane mocking
- More convenient CLI functionality

mute:
## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [x] IE11 (Win 7)
  - [x] Edge (Win 10)

Fixes #612


src/components/Resizer/Resizer.jsx
```diff
@@ -12,26 +12,36 @@ const {
...
- * containing element. This component doesn't take normal children. It expects
- * you to pass a single function for children. It will then call that function
- * with new `width` and `height` values if the container size changes.
+ * Contains the base functionality of `Resizer` but allows for injection of
+ * `listenTo` and `removeListener` function dependencies to improve
+ * testability.
  */
-const Resizer = createClass({
-	displayName: 'Resizer',
+export const ResizerBase = createClass({
```
mute: I added this base component just to make it more convenient to mock the dependencies of `Resizer`. It provides a way of injecting the dependencies for `Resizer` to obviate the need for using Rewire to change internal values.

src/util/logger.js
```diff
@@ -44,7 +44,6 @@ function resetOnce(key) {
 export function checkIsDev() {
 	return isDevMode &&
 		typeof window !== 'undefined' &&
-		!isNode &&
```
mute: I removed this because of the tests for `logger`. They were failing the `isDevMode` check because of running in node.

I don't think we actually need it here but I don't know why it was added originally.

src/util/component-types.js
```diff
@@ -24,6 +25,7 @@ export function createClass(definition={}) {
 			selectors,
 			propName,
 		},
+		propTypes: _.assign({}, propTypes, _.mapValues(definition.components, () => React.PropTypes.any)),
```
mute: This change adds anything defined in `components` to the `propTypes` list for the component, to remove warnings when valid Child Components are being passed in as props.

src/components/AxisLabel/AxisLabel.spec.jsx
```diff
@@ -102,7 +107,11 @@ describe('AxisLabel', () => {
 		describe('label', () => {
```
sodiumjoe: i feel like snapshots obviate a lot of this kind of test now

mute: That's a good point -- as part of this PR I didn't go through and convert any existing tests over to using snapshots, or remove tests that are replaced with Example snapshots.

I'll make an effort to do some of that now, but it's not likely that I'm going to catch everything.

src/components/RadioGroup/examples/default.jsx
```diff
@@ -8,7 +8,7 @@ const style = {
 export default React.createClass({
 	render() {
 		return (
-			<RadioGroup>
+			<RadioGroup name='name'>
```
ogupte: This kind of changes the nature of the example a bit. One main feature is that you don't have to provide an arbitrary name. but now all the examples include it so that the tests will be consistent.

src/components/Resizer/Resizer.util.js
```diff
@@ -0,0 +1,3 @@
+import elementResizeDetectorMaker from 'element-resize-detector';
+
+export const erd = elementResizeDetectorMaker({strategy: 'scroll'});
```
ogupte: Should this const be more descriptive?

src/components/SearchableSelect/SearchableSelect.jsx
```diff
@@ -149,7 +149,7 @@ const SearchableSelect = createClass({
 			isDisabled: false,
 			isLoading: false,
 			optionFilter: this.defaultOptionFilter,
-			searchText: null,
+			searchText: '',
```
ogupte: The issue here is actually with SearchField, see #614

src/components/Validation/Validation.spec.jsx
```diff
@@ -5,8 +5,11 @@ import { shallow } from 'enzyme';
 import Validation from './Validation';
 
 describe('Validation', () => {
-
-	common(Validation);
+	common(Validation, {
+		getDefaultProps: () => ({
+			children: <span>foo</span>,
+		}),
+	});
```
ogupte: i don't understand why this is needed.

src/components/RadioGroup/examples/default.jsx
mute: That's a good point -- what do you think about using [_.uniqueId](https://lodash.com/docs/4.16.6#uniqueId) rather than our `random()` to generate these IDs? That would let me keep the example the same but also give consistent unique identifiers.

src/components/Resizer/Resizer.util.js
mute: 👍

src/components/RadioGroup/examples/default.jsx
ogupte: i agree that it's better than a random

src/components/SearchableSelect/SearchableSelect.jsx
mute: Ah okay, thank you -- I'll leave this unchanged then.

src/components/AxisLabel/AxisLabel.spec.jsx
sodiumjoe: yeah, we can probably do it as we go

PR #635, opened 2016-11-28
Adds snapshot directories to .npmignore

mute:
## PR Checklist


sodiumjoe:
I wonder if we should enumerate files to include in the `files` property of `package.json` instead

PR #652, opened 2016-12-06
Adds new `SplitButton` component.

### LINKED ISSUE #630: SplitButton

mute:
## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [x] IE11 (Win 7)
  - [x] Edge (Win 10)

This PR also includes a fix for an example snapshot bug which was preventing hybrid component examples from being evaluated correctly.

Fixes #630.

Docs here: http://docspot.devnxs.net/projects/lucid/630-split-button/#/components/SplitButton


sodiumjoe:
why did unrelated snapshots change?

src/components/SplitButton/SplitButton.jsx
```diff
@@ -0,0 +1,147 @@
...
+		return (
+			<DropMenu
+				{...dropMenuProps}
+				{...omitProps(passThroughs, SplitButton)}
+				className={cx('&', className)}
+				onSelect={this.handleSelect}
+			>
+				<DropMenu.Control>
+					<ButtonGroup>
+						<Button
```
sodiumjoe: I wonder if the primary button should be explicit.

src/components/SplitButton/SplitButton.jsx
```diff
@@ -0,0 +1,147 @@
...
+			DropMenu: dropMenuProps,
+			...passThroughs
+		} = this.props;
+
+		const {
+			isExpanded,
+		} = dropMenuProps;
+
+		const buttonChildProps = _.map(filterTypes(this.props.children, SplitButton.Button), 'props');
+		const primaryButtonProps = _.first(buttonChildProps);
```
sodiumjoe: this could be
```
const [primaryButtonProps, ...buttonChildProps] = _.map(filterTypes());
```
to avoid the `_.slice` below

mute:
I found an issue with example snapshotting for hybridComponents while working on this that meant that none of the examples for `SplitButton` would end up getting tested correctly. With the fix for that, it meant that other hybridComponents would also be getting snapshotted properly, which resulted in the additional snapshots.

src/components/SplitButton/SplitButton.jsx
sodiumjoe: actually, seems a little weird to me that the `Option`s are `Button`s.

sodiumjoe: maybe should be `SplitButton.Button` and `SplitButton.Option`

sodiumjoe:
`(unknown) Warning: Unknown props `isActive`, `hasOnlyIcon` on <div> tag.` on the examples page

sodiumjoe:
the `link` one on the examples page has some funny behavior, too

1.
screenshot 2016-12-06 09 51 32

2. the drop menu doesn't close after selecting an option?

src/components/SplitButton/SplitButton.less
```diff
@@ -0,0 +1,12 @@
+.lucid-SplitButton {
+	.lucid-CaretIcon {
```
sodiumjoe: i think we should add a `SplitButton` specific class and use that

src/util/generic-tests.jsx
```diff
@@ -121,7 +121,7 @@ export function common(Component, {
 					if (shallowExample.is(Component.displayName)) {
 						expect(shallowToJson(shallow(<Component {...shallowExample.props()} />))).toMatchSnapshot();
 					} else {
-						shallowExample.find(Component).forEach((example) => {
+						shallowExample.find(Component.displayName).forEach((example) => {
```
sodiumjoe: weird that the component reference doesn't work

mute: It's only an issue for hybridComponents, because the instance that gets passed in to `common` tests generally is the Dumb version rather than the version from `index`

src/components/SplitButton/SplitButton.less
mute: 👌

src/components/SplitButton/SplitButton.jsx
```diff
@@ -0,0 +1,215 @@
...
+		 * first `Button` will be used as the Primary button, while all others will
+		 * be rendered within the `DropMenu` below.
+		 */
+		Button: createClass({
+			displayName: 'SplitButton.Button',
+			propTypes: {
+				/**
+				 * The children to render within the `Button`.
+				 */
+				children: any,
```
ogupte: do you really want `any` here, or is `node` a little more appropriate? even tho both are kind of ambiguous.

src/components/SplitButton/SplitButton.jsx
```diff
@@ -0,0 +1,215 @@
...
+				{...omitProps(passThroughs, SplitButton)}
+				direction={direction}
+				className={cx('&', className)}
+				onSelect={this.handleSelect}
+			>
+				<DropMenu.Control>
+					<ButtonGroup>
+						<Button
+							{...primaryButtonProps}
+							className={cx('&-Primary-Button')}
```
ogupte: i think we should only capitalize parts of the className when it refers to a component type, `Button` makes sense but `primary` should be lowercase. Also i think the type should be front-loaded.

mute: So `Button-primary`?

src/components/SplitButton/SplitButton.jsx
mute: `node`

mute: Fixing this should be `node` problem!

Issue #651, opened 2016-12-05
Fix style scoping issue with Panel


njlg:
Nested a panel inside a panel means some of the parent's styles are inherited by the child.

mute:
Fixed by #653

PR #653, opened 2016-12-06
Panel: Add new Panel prop `isScrollable`

ogupte:
## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [x] IE11 (Win 7)
  - [x] Edge (Win 10)


mute:
Is this fixing #651 also?

ogupte:
re: https://github.com/appnexus/lucid/pull/653#issuecomment-265506596

Yea in the process of cleaning up the LESS, it happens to fix #651. I didn't intend to fix it but good catch.

ogupte:
re: https://github.com/appnexus/lucid/pull/653#issuecomment-265506596

I don't think it's very valuable to have an example with `isScrollable={false}`. It only exists as a prop for backward compat reasons, but in the next major, all overflowing content should be scrollable in Panel, and this prop should be removed.

src/components/Panel/Panel.less
```diff
@@ -6,23 +6,36 @@
 	background-color: @color-white;
 	border: @border-standardBorder;
 
-	&-has-margin {
+	.lucid-Panel-has-margin& {
```
jondlm: I think we typically do `&&-has-margin`

src/components/Panel/Panel.less
```diff
@@ -6,23 +6,36 @@
...
+	& > &-Header {
 		padding: @size-standard;
 		border-bottom: @border-lightBorder;
 	}
 
-	&-is-not-gutterless &-content {
-		padding: @size-standard;
-		overflow: auto;
+	& > &-content {
+		.lucid-Panel-is-not-gutterless& {
```
jondlm: This rule is confusing as well, it equates to this at the root level: 

```less
&&-is-not-gutterless > &-content {
    overflow: auto;
}
```

ogupte: This is following the LESS docs to select for parent classes (http://lesscss.org/features/#parent-selectors-feature-changing-selector-order).

For example we could have three different selectors with `&-content` in three different places, which resembles  vanilla CSS more closely:

```
& > &-content {
  //default styles
}
&&-is-not-gutterless > &-content {...}
&&-is-scrollable > &-content {...}
```

Or, we can have one section for content with defaults styles with a parent class selector to extend them:

```
& > &-content {
  //default content styles

  .lucid-Panel-is-not-gutterless& {...}
  .lucid-Panel-is-scrollable& {...}
}
```

Both approaches produce the same CSS, but the latter is better for codeorg by making use of available LESS features. I've used this pattern in other components in this repo, and feel we should be making more use of it.

src/components/Panel/Panel.less
ogupte: will fix

src/components/Panel/Panel.less
jondlm: Ahh. I didn't know about that feature. Sounds good!

PR #655, opened 2016-12-07
Added `quote-props` lint rule with fixes

tsantef:
## PR Checklist


ogupte:
I don't think this linting rule should be enforced. It's useful to quote object properties to convey that object keys might be dynamic, like when mocking data for table tests.

PR #674, opened 2016-12-16
Adding new props to DropMenu

### LINKED ISSUE #673: Expose ContextMenu props through DropMenu
I'm using `DropMenu` on a project and I need to be able to set some props on the underlying `ContextMenu`.

jondlm:
## PR Checklist

http://docspot.devnxs.net/projects/lucid/673-expose-contextmenu-props/#/components/DropMenu

See the `No Wrapping` example

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [x] IE11 (Win 7)
  - [x] Edge (Win 10)

Fixes #673


src/components/DropMenu/DropMenu.jsx
```diff
@@ -89,6 +94,17 @@ const DropMenu = createClass({
 				 * hides the `Option` from the list.
 				 */
 				isHidden: bool,
+				/**
+				 * controls wrapping of the text.
+				 */
+				isWrapped: bool,
+			},
+			getDefaultProps() {
```
jondlm: These were missing default props so I added them and updated snapshots.

src/components/DropMenu/DropMenu.jsx
```diff
@@ -429,6 +451,7 @@ const DropMenu = createClass({
 						'&-Option-is-selected': isSelected,
 						'&-Option-is-disabled': isDisabled,
 						'&-Option-is-null': _.isNull(optionIndex),
+						'&-Option-is-wrapped': isWrapped,
```
mute: I'm getting confused a bit by this prop name -- it's called `isWrapped` but it actually applies `nowrap` if it is `true` and then the example for using it is called `no-wrapping`. Is that correct?

jondlm: You're totally right, I messed up the naming.

PR #688, opened 2017-01-18
create scroll listener on mount instead of exported singleton

jhsu:
the existing module exports an instance of `element-resize-detector` which tries to bind event listeners on `document.body` which may not be available at instantiation.

## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [ ] IE11 (Win 7)
  - [ ] Edge (Win 10)

### DIFF
--- a/src/components/Resizer/Resizer.spec.jsx
+++ b/src/components/Resizer/Resizer.spec.jsx
@@ -1,17 +1,21 @@
-jest.mock('./Resizer.util');
+jest.mock('element-resize-detector', () => {
+	return function() {
+		return {
+			listenTo: jest.fn((_element, handleResize) => {
+				handleResize({
+					offsetWidth: 50,
+					offsetHeight: 100,
+				});
+			}),
+			removeListener: jest.fn(),
+		};
+	};
+});
 import _ from 'lodash';
 import React from 'react';
 import { common } from '../../util/generic-tests';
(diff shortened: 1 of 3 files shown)


jhsu:
this creates a new instance which probably creates a new listener per instance of the resizer.

one possibility that I know react-waypoint uses is creates an event listener on the parent's dom node so that any sibling checks the parent for a listener before creating its own.

sodiumjoe:
hm, generally there are at most a few resizing elements, right? Seems ok to me

src/components/Resizer/Resizer.spec.jsx
```diff
@@ -1,17 +1,21 @@
...
+			removeListener: jest.fn(),
+		};
+	};
+});
 import _ from 'lodash';
 import React from 'react';
 import { common } from '../../util/generic-tests';
 import { mount } from 'enzyme';
 import Resizer from './Resizer';
-import { elementResizeDetector } from './Resizer.util';
```
sodiumjoe: wasn't there some reason we had to do this in the first place @mute ?

mute: We only pulled this out to a `util` to maintain the singleton while making testing a little easier.

If we're okay with it not being a singleton, we don't need it anymore.

PR #694, opened 2017-01-26
Path imports

### LINKED ISSUE #693: Improve path imports
Our bundle is pretty big, and tree shaking hasn't quite panned out the way we'd hoped. In my testing it didn't work nearly as well as path imports. There's [a section of the rollup's docs](https://github.com/rollup/rollup/wiki/Troubleshooting#tree-shaking-doesnt-seem-to-be-working) that explains it quite well.

That being said, we'd like to do a few things to reduce our bundle size and make it easier for consumers to do the same:

1. [ ] use `lodash` path imports inside lucid
1. [ ] modify our build process to read the `index` file and produce a set of flat files that consumers can use path imports. E.g. `import Button from 'lucid-ui/Button';`

jondlm:
Fixes #693. This PR introduces a set of build tasks to support path imports. It also includes the lodash babel plugin to reduce our dependency on lodash.

The lodash changes brought our browser build from `2.39 MB` to `2.17 MB`. I wasn't expecting to see much shrinking there. The benefit is really to people who import the components using the path import approach.

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] IE11 (Win 7)
  - [ ] Edge (Win 10)


src/components/Accordion/__snapshots__/Accordion.spec.jsx.snap
```diff
@@ -13,7 +13,7 @@ exports[`Accordion [common] example testing should match snapshot(s) for 1.basic
     className="lucid-Accordion-Item"
     isExpanded={false}
     onToggle={[Function]}>
-    Raymond \"Ray\" Stantz, Ph.D. is a fictional character from the Ghostbusters franchise. He appears in the films Ghostbusters, Ghostbusters II, Casper, and the animated television series The Real Ghostbusters. He was portrayed by Dan Aykroyd in both live action films, and voiced by Frank Welker in the animated series. He is a member of the Ghostbusters and one of the three doctors of parapsychology, along with Dr. Peter Venkman and Dr. Egon Spengler.
```
jondlm: I'm not sure why these changed.

package.json
```diff
@@ -13,11 +13,11 @@
     "dev": "gulp docs-generate && webpack-dev-server --inline --content-base src/docs/",
     "docs-upload-gh": "scripts/docs-upload-gh.sh",
     "docs-upload": "gulp docs-upload",
-    "jest": "TZ=UTC jest",
+    "jest": "NODE_ENV=test TZ=UTC jest",
```
jondlm: This helps key babel to _not_ use the lodash plugin during testing

package.json
```diff
@@ -50,14 +50,16 @@
     "reselect": "^2.5.1"
   },
   "devDependencies": {
-    "anx-docgen": "^1.0.2",
+    "anx-docgen": "^1.1.0",
```
jondlm: I had to make a change to anx-docgen to support exporting hybrid components.

package.json
```diff
@@ -50,14 +50,16 @@
...
     "autoprefixer": "^6.3.3",
     "babel-core": "^6.4.5",
     "babel-loader": "^6.2.1",
+    "babel-plugin-lodash": "^3.2.11",
     "babel-preset-airbnb": "^2.0.0",
     "babel-preset-es2015": "^6.3.13",
     "babel-preset-react": "^6.3.13",
     "babel-preset-stage-2": "^6.5.0",
+    "babylon": "^6.0.0",
```
jondlm: This is already a nested dep of babel, so it shouldn't add to our `yarn` time.

src/components/DragCaptureZone/examples/interactive.jsx
```diff
@@ -41,26 +41,21 @@ export default React.createClass({
 					overflow: 'auto',
 					width: 600,
 				}}>
-					{
-						_.chain(this.state.events)
```
jondlm: Had to remove all usage of `chain`. In most cases I just went with nesting, but for longer chains I used `lodash/fp/flow`

gulp/build.js
```diff
@@ -30,4 +65,65 @@ module.exports = {
...
+
+	/**
+	 * This is a task that reads our `index.js` and statically analyzes it for
+	 * imports/exports. Based on that, it writes a single file for each of the
+	 * exports to the root directory. The whole purpose is to make path imports,
+	 * i.e. `import Button from 'lucid-ui/Button'`, easy.
+	 *
+	 * Because there is a potential for file system collisions, we have to
+	 * perform a check to make sure none of the files we're going to write
+	 * already exist. If they do, we'll bail out and log a waring.
```
sodiumjoe: typo

> waring

src/components/PieChart/PieChart.jsx
```diff
@@ -1,12 +1,13 @@
-import _ from 'lodash';
+import _  from 'lodash';
```
sodiumjoe: extra space

src/docs/containers/icons.jsx
```diff
@@ -1,20 +1,21 @@
...
+	(x) => _.reduce(x, (acc, val, key) => (
 		acc.concat(_.endsWith(key, 'Icon') && key !== 'Icon' ? [[key, val]] : [])
-	), [])
-	.filter(([, Icon]) => !Icon._lucidIsPrivate)
-	.sortBy(([name]) => name)
-	.value();
+	), []),
+	(x) => _.filter(x, ([, Icon]) => !Icon._lucidIsPrivate),
+	(x) => _.sortBy(x, ([name]) => name)
+)(components);
```
sodiumjoe: I don't think the `fp` version of `flow` is doing anything for us here. You could just use `_.flow`. The benefit of using `fp/flow` doesn't really appear until you use all the `fp` methods, so it would look like:

```
flow(
  reduce((acc, val, key) => acc.concat(_.endsWith(key, 'Icon') && key !== 'Icon' ? [[key, val]] : []),
  filter([, Icon]) => !Icon._lucidIsPrivate),
  sortBy([name]) => name)
)(components)
```

sodiumjoe: docs code (code that's not included in the public api) might be a good place for us to try out fp, actually

src/components/Accordion/__snapshots__/Accordion.spec.jsx.snap
sodiumjoe: looks like just escapes got removed from the output, maybe a minor bump of jest or something?

src/docs/containers/icons.jsx
jondlm: That's a good point, I didn't even realize `_.flow` was a thing.

gulp/build.js
```diff
@@ -30,4 +65,65 @@ module.exports = {
...
+					}
+					return resolve();
+				});
+			});
+		}));
+
+		const writeFilesPromise = Promise.all(_.map(allExports, (exportName) => {
+			return new Promise((resolve, reject) => {
+				const exportCode = `export * from '${importMap[exportName]}';`;
+				// for some reason this doesn't pickup our .babelrc even though it's supposed to
```
mute: Huh, that's really odd.

package.json
```diff
@@ -13,11 +13,11 @@
     "dev": "gulp docs-generate && webpack-dev-server --inline --content-base src/docs/",
     "docs-upload-gh": "scripts/docs-upload-gh.sh",
     "docs-upload": "gulp docs-upload",
-    "jest": "TZ=UTC jest",
+    "jest": "NODE_ENV=test TZ=UTC jest",
     "lint": "eslint . --ext .js,.jsx",
     "lint-fix": "eslint . --ext .js,.jsx --fix",
     "preversion": "rm -rf dist && npm run lint && gulp preversion && npm run build-bundle",
-    "postversion": "gulp docs-upload",
+    "postversion": "gulp build-js-clean-path-imports && gulp docs-upload",
```
mute: It might be nice to add an `npm clean` script here also that runs this, too.

gulp/build.js
jondlm: It doesn't really matter since each file is only a single line of code that's an `export`

PR #706, opened 2017-02-03
Fix path imports

jondlm:
## PR Checklist

  - [x] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] IE11 (Win 7)
  - [ ] Edge (Win 10)

### DIFF
--- a/gulp/build.js
+++ b/gulp/build.js
@@ -104,9 +104,17 @@ module.exports = {
 				const specifierPath = importMap[exportName].path;
 				const specifierType = importMap[exportName].type;
 
-				// Sorry, this is a little hard to read
-				const exportCode = specifierType === 'ImportSpecifier' ? `export * from '${specifierPath}';`
-					: specifierType === 'ImportDefaultSpecifier' ? `export * from '${specifierPath}';`
+				// Sorry, this is a little hard to read, it was hard for me to figure
+				// out :facepalm:
+				const exportCode = specifierType === 'ImportSpecifier' ? `
+import def from '${specifierPath}';
+export default def;
+export * from '${specifierPath}';
+`
+					: specifierType === 'ImportDefaultSpecifier' ? `
+import def from '${specifierPath}';
+export default def;
+`
 					: specifierType === 'ImportNamespaceSpecifier' ? `
 import * as def from '${specifierPath}';
 export default def;
--- a/gulp/build.js
+++ b/gulp/build.js
@@ -28,10 +28,15 @@ function imports() {
 		// root directory instead of `dist`
 		const newPath = isPath ? './' + path.join(CONFIG.BUILD_DIR, moduleId) : null;
 
-		return _.map(node.specifiers, ((specifier) => ([
-			specifier.local.name,
-			newPath || moduleId,
-		])));
+		return _.map(node.specifiers, ((specifier) => {
+			return [
+				specifier.local.name,
+				{
+					type: specifier.type,
+					path: newPath || moduleId,
+				},
+			];
+		}));
 	});
 }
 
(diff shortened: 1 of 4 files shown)


gulp/build.js
```diff
@@ -96,8 +101,28 @@ module.exports = {
...
 		const writeFilesPromise = Promise.all(_.map(allExports, (exportName) => {
 			return new Promise((resolve, reject) => {
-				const exportCode = `export * from '${importMap[exportName]}';`;
-				// for some reason this doesn't pickup our .babelrc even though it's supposed to
+				const specifierPath = importMap[exportName].path;
+				const specifierType = importMap[exportName].type;
+
+				// Sorry, this is a little hard to read, it was hard for me to figure
+				// out :facepalm:
+				const exportCode = specifierType === 'ImportSpecifier' ? `
```
jondlm: There are three kinds of imports, this covers all three of those cases and most importantly ensures there's a default export in each case.

jondlm: I dunno, the switch is pretty ugly too: ```				let exportCode = '';

				switch (specifierType) {
				case 'ImportSpecifier':
					exportCode = `
import { ${exportName} } from '${specifierPath}';
export default ${exportName};
export * from '${specifierPath}';
`;
					break;
				case 'ImportDefaultSpecifier':
					exportCode = `
import def from '${specifierPath}';
export default def;
`;
					break;
				case 'ImportNamespaceSpecifier':
					exportCode = `
import * as def from '${specifierPath}';
export default def;
export * from '${specifierPath}'
`;
				}
```

File: src/util/dom-helpers.spec.js [hunk 2]
```diff
@@ -69,7 +69,9 @@ describe('#scrollParentTo', () => {
 		assert.equal(parent.scrollTop, 5);
 	});
 
-	it('should align to bottom if the bottom of the node is below the fold', () => {
+	// This test cannot be run anymore because `clientHeight` cannot be mutated
+	// and I wasn't able to figure out how to set it
+	it.skip('should align to bottom if the bottom of the node is below the fold', () => {
```
sodiumjoe: ☝️

sodiumjoe: `clientHeight` can't be mutated in jsdom anymore? Is this from a jest update?

gulp/build.js
```diff
@@ -96,8 +101,28 @@ module.exports = {
...
 		const writeFilesPromise = Promise.all(_.map(allExports, (exportName) => {
 			return new Promise((resolve, reject) => {
-				const exportCode = `export * from '${importMap[exportName]}';`;
-				// for some reason this doesn't pickup our .babelrc even though it's supposed to
+				const specifierPath = importMap[exportName].path;
+				const specifierType = importMap[exportName].type;
+
+				// Sorry, this is a little hard to read, it's effectively a switch
+				// statement on specifierType
+				const exportCode = specifierType === 'ImportSpecifier' ? `
```
sodiumjoe: i think it would be cleaner with our usual
```
condition
? then
: else
```
format, esp. if combined with https://github.com/declandewet/common-tags#stripindent

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "lucid-ui",
-  "version": "2.13.0",
+  "version": "2.13.1-path-imports.4",
```
sodiumjoe: remove?

src/util/dom-helpers.spec.js
jondlm: I believe so, I couldn't find an alternative either.

sodiumjoe: it was never a great test to begin with... i vote for delete

PR #714, opened 2017-02-14
Always omit "callbackId" prop

jondlm:
## PR Checklist


src/util/component-types.js
```diff
@@ -94,8 +94,12 @@ export function getFirst(props, types, defaultValue) {
...
-// and any extra keys given in third argument
+// Omit props defined in propTypes of the given type and any extra keys given
+// in third argument
+//
+// We also include a "magic" prop that's always excluded called `callbackId`.
+// That prop can be used to identify a component in a list without having to
+// create extra closures.
 export function omitProps(props, type, keys = []) {
-	return _.omit(props, _.keys(type.propTypes).concat(keys).concat(['initialState']));
+	return _.omit(props, _.keys(type.propTypes).concat(keys).concat(['initialState', 'callbackId']));
```
mute: We talked about this a little bit, but if our goal is for this prop to cover all closure cases, I think it would be nice if the name were more general rather than tied to `id`.

Ideas: `callbackProp`, `callbackProps`, `callbackValue`.

jondlm: After thinking a little bit about this, it seems like the 95% case is going to be an `id` and we should name it accordingly. It also encourages people to keep use something simple that's uniquely identifiable.

jondlm:
Found a bug w/ `RadioGroup` (thanks tests!) which uses `callbackId` internally and excepts it to be passed through a couple layers of lucid components.

mute:
It's probably bad that my initial instinct was, "The tests must be wrong"...

PR #720, opened 2017-02-21
Jest 19

mute:
## PR Checklist


src/components/Accordion/__snapshots__/Accordion.spec.jsx.snap
```diff
@@ -1,102 +1,118 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
sodiumjoe: weird that this is a shortened url and a google one

jondlm: It's so they can use code search to see how many snapshots people are writing :trollface:

sodiumjoe: hope they're not using hound

PR #722, opened 2017-02-22
Fix SingleSelect in RadioGroup bug

### LINKED ISSUE #662: SingleSelect does not render properly inside RadioGroup.Label
In order to show/hide content underneath radio buttons (as they are selected), I tried using the RadioGroup component and placed selection-specific content inside the RadioGroup.Label component (with a conditional wrapper to specify whether or not it should be shown). Ex: 

```
<RadioGroup.RadioButton>
    <RadioGroup.Label>
        <span>{label}</span>
        { 
            isOptionOneSelected 
            ? <OptionOneContent />
            : null
        }
    </RadioGroup.Label>
</RadioGroup.RadioButton>
```

This generally works, but when I try to place a SingleSelect component inside RadioGroup.Label, the menu does not render when the SingleSelect is clicked. It works fine outside of RadioGroup, however. It doesn't work inside the RadioButtonLabeled component either.

jondlm:
Fixes #662

Detects if a body click was a `<label>` and doesn't trigger the on click
out handler if that's the case.

## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [x] IE11 (Win 7)
  - [x] Edge (Win 10)

### DIFF
--- a/src/components/ContextMenu/ContextMenu.jsx
+++ b/src/components/ContextMenu/ContextMenu.jsx
@@ -2,7 +2,7 @@ import React from 'react';
 import _ from 'lodash';
 import Portal from '../Portal/Portal';
 import { createClass, getFirst, omitProps } from '../../util/component-types';
-import { getAbsoluteBoundingClientRect } from '../../util/dom-helpers';
+import { getAbsoluteBoundingClientRect, sharesAncestor } from '../../util/dom-helpers';
 import { lucidClassNames } from '../../util/style-helpers';
 
 const cx = lucidClassNames.bind('&-ContextMenu');
@@ -133,7 +133,6 @@ const ContextMenu = createClass({
 	},
 
 	componentDidMount() {
-
 		_.defer(() => this.alignFlyOut());
 		this.updateTargetRectangleIntervalId = setInterval(() => {
 			if (this.props.isExpanded) {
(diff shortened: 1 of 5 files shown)


src/components/ContextMenu/ContextMenu.jsx
```diff
@@ -133,27 +133,46 @@ const ContextMenu = createClass({
...
+			},
+			refs: {
+				flyOutPortal,
+				target,
+			},
+		} = this;
+
+		if (onClickOut && flyOutPortal) {
+			const flyOutEl = flyOutPortal.portalElement.firstChild;
+			const wasALabelClick = event.target.nodeName === 'INPUT' && sharesAncestor(event.target, target, 'LABEL');
```
jondlm: I learned that `<label>`s fire a real click event with a target of the corresponding `INPUT`. This line checks to see if the event fired relates to an input and shares a label ancestor with the context menu target.

src/util/dom-helpers.js
```diff
@@ -35,3 +35,30 @@ export function dispatchDOMEvent(node, eventName, assignedEventProps) {
...
+
+/**
+ * sharesAncestor
+ *
+ * Recursively looks at `node` and its parents for `findNodeName` and makes
+ * sure it contains `siblingNode`.
+ *
+ * @param {element} node - dom node to check if any of its ancestors are a `<label>`
+ * @param {element} siblingNode - dom node to see if it shares an ancestor
+ * @param {string} findNodeName - dom node name, should be uppercased, e.g. `LABEL` or `SPAN`
```
ogupte: `findNodeName` sounds like a function, maybe rename to just `nodeName`

src/util/dom-helpers.js
```diff
@@ -35,3 +35,30 @@ export function dispatchDOMEvent(node, eventName, assignedEventProps) {
...
+export function sharesAncestor(node, siblingNode, findNodeName) {
+	const nodeName = _.get(node, 'nodeName');
+	const parentNode = _.get(node, 'parentNode');
+
+	if (nodeName === findNodeName) {
+		return node.contains(siblingNode);
+	}
+
+	if (parentNode) {
+		return false || sharesAncestor(parentNode, siblingNode, findNodeName);
```
mute: Seems like the `false ||` might be redundant here -- the recursive call should return either false or the boolean from `node.contains` right?

PR #730, opened 2017-03-03
Shifts `onClick` and `onMouseMove` to take precedence over option props.

### LINKED ISSUE #729: `DropMenu` still fires `onClick` for `isDisabled` child `Option`s.
If the `Option` has the `isDisabled` prop, it shouldn't trigger an `onClick` handler.

mute:
## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [x] Edge (Win 10)

Fixes #729, #730


mute:
I'm not sure if we want to make this change or not -- `onClick` isn't technically a supported prop on `DropMenu.Option` but it seems possible that a consumer may have been depending on it.

PR #735, opened 2017-03-06
Add lockfile, fix tests

jondlm:
## PR Checklist


.gitignore
```diff
@@ -6,9 +6,6 @@ src/docs/docgen.json
 npm-debug.log*
 node_modules
 
-# Yarn
```
jondlm: As core contributors we should be doing `yarn upgrade` and be checking in lockfile changes.

.travis.yml
```diff
@@ -1,14 +1,5 @@
 before_install:
-  # Repo for Yarn
```
jondlm: All this stuff isn't needed anymore since it's built into travis now and we are checking in our yarn.lock file.

src/components/CalendarMonth/__snapshots__/CalendarMonth.spec.jsx.snap
```diff
@@ -55,6 +55,12 @@ exports[`CalendarMonth common tests [common] example testing should match snapsh
   fixedWeeks={false}
   from={null}
   initialMonth={2016-02-17T22:25:23.000Z}
+  labels={
```
jondlm: Downstream change in react day picker I think

package.json
```diff
@@ -13,7 +13,7 @@
     "dev": "gulp docs-generate && webpack-dev-server --inline --hot --content-base src/docs/",
     "docs-upload-gh": "scripts/docs-upload-gh.sh",
     "docs-upload": "gulp docs-upload",
-    "jest": "NODE_ENV=test TZ=UTC jest",
+    "jest": "NODE_ENV=test TZ=UTC jest --runInBand",
```
jondlm: For some very strange reason, this flag was needed otherwise the tests would hang on travis. :man_shrugging:

chrisdopuch: It would be good if there was a way to only run in band on CI mode, and continue to multi-thread when running it locally, since they run faster that way.

jondlm: Totally, we usually use the `npm run watch` command to run tests while developing and that isn't affected by this, so it's a workable way to use multiple threads.

PR #737, opened 2017-03-13
DateSelect component style and props improvements

ogupte:
## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [x] IE11 (Win 7)
  - [x] Edge (Win 10)


src/components/DateSelect/DateSelect.jsx
```diff
@@ -201,6 +216,30 @@ const DateSelect = createClass({
...
+			const rootElement = ReactDOM.findDOMNode(this.rootRef);
+			const {
+				width,
+				height,
+			} = rootElement.getBoundingClientRect();
+			const navButtonsWidth = NAV_BUTTON_SIZE * 2;
+			const oneMonthShownWidth = (width - navButtonsWidth) / monthsShown + navButtonsWidth;
+			const size = Math.sqrt(oneMonthShownWidth * height);
+			const relativeFontSize = Math.round(size / 24);
+			const relativeMinWidth = (((width - navButtonsWidth) / monthsShown) * 10.1075 / relativeFontSize) * monthsShown + navButtonsWidth;
```
jondlm: `10.1075`? Did it just look good in your testing?

src/components/DateSelect/DateSelect.jsx
```diff
@@ -201,6 +216,30 @@ const DateSelect = createClass({
...
+
+			rootElement.style.fontSize = `${relativeFontSize}px`;
```
jondlm: So this'll only change once at the component mount, is that an issue for the prop changing?

src/components/DateSelect/DateSelect.jsx
ogupte: yeah i calculated this as the ratio that the default font-size with the default size of the DateSelect. so in order to scale the font-size, i use the same ratio here

src/components/DateSelect/DateSelect.jsx
ogupte: it's documented in the prop as only affecting the _initial_ font-size so i think it's ok. i'd rather not have to recompute onresize and on props change

jondlm: Cool, as long as we call it out I'm fine with that

PR #742, opened 2017-03-14
Add support for DropMenu.Option onClick handler that is passed through

ogupte:
## PR Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] IE11 (Win 7)
  - [ ] Edge (Win 10)

### DIFF
--- a/src/components/DropMenu/DropMenu.jsx
+++ b/src/components/DropMenu/DropMenu.jsx
@@ -414,6 +414,7 @@ const DropMenu = createClass({
 
 		if (!optionProps.isDisabled && focusedIndex !== optionIndex) {
 			onFocusOption(optionIndex, { props: optionProps, event });
+			optionProps.onClick({{ props: optionProps, event }});
 		}
 	},
 


src/components/DropMenu/DropMenu.jsx
```diff
@@ -414,6 +414,7 @@ const DropMenu = createClass({
 
 		if (!optionProps.isDisabled && focusedIndex !== optionIndex) {
 			onFocusOption(optionIndex, { props: optionProps, event });
+			optionProps.onClick({{ props: optionProps, event }});
```
jondlm: Does this break when there is no `onClick`?

chrisdopuch: ➕ 1 you could add a `_.noop` to the `getDefaultProps`

mute: It seems weird to fire two events for a single action; should it be a precedence thing? e.g.:
`if optionProps.onClick is set` -> call that function
`else -> ` call `onSelect` function

ogupte:
this PR was for illustration and discussion purposes, closing it now.

PR #750, opened 2017-04-11
Switches to using prop-types library.

mute:
## PR Checklist


mute:
Some other dependencies are still causing a dramatic number of warnings during tests. I didn't track down what they were, but this is probably going to become a problem for consumers of lucid as well.

PR #753, opened 2017-04-12
703 optimizations

### LINKED ISSUE #703: Redux utils mergeProps doesn't maintain references correctly
Through a bunch of testing on one of our internal components, we found that our redux util's `mergeProps` is breaking reselect selectors. We think we need to write our own merge that maintains referential integrity without mutation.

- [ ] Fix merge props so that it doesn't create new references
- [ ] Double check and make sure that `omitProps` isn't adding new references either

sodiumjoe:
## PR Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] IE11 (Win 7)
  - [ ] Edge (Win 10)

fixes #703 (sorta, i.e. it does make mergeProps maintain references, but not omitProps. omitProps is trickier, because the component calling it would have to make sure to use static function references instead of anonymous functions. But I think the general memoization of selectors sidesteps the issue in a lot of ways since it will avoid even getting to a component's render if its props haven't changed)


src/util/state-management.js
```diff
@@ -77,19 +78,29 @@ export function getStatefulPropsContext(reducers, { getState, setState }) {
  * Generates a root selector from a tree of selectors
  * @param {Object} selectors - a tree of selectors
  * @returns {function} root selector that when called with state, calls each of
- * the selectors in the tree with the state local to that selector
+ * the selectors in the tree with the state local to that selector.
+ *
+ * This function is memoized because it's recursive, and we want it to reuse
+ * the functions created in the recursive reduce (because those functions are
```
chrisdopuch: I'm not sure I'm following the part in parenthesis. Which other functions are also memoized? The selector functions? Is that something we do, or a part of how reselect works?

sodiumjoe: yeah, reselect basically memoizes with a cache one 1

src/util/redux.js
```diff
@@ -261,9 +261,9 @@ function bindActionCreatorTree(actionCreatorTree, dispatch, path = []) {
  * @param {Object} ownProps
  * @return {Object}
  */
-function mergeProps(state, dispatchTree, ownProps) {
-	return _.merge({}, state, dispatchTree, ownProps);
-}
+const mergeProps = _.memoize((state, dispatchTree, ownProps) => {
```
jondlm: I really hope we don't mutate state anywhere

sodiumjoe: 😬

PR #808, opened 2017-08-09
Anxr 824 - Render minRows filler rows when data is empty

chrisdopuch:
## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [x] Edge (Win 10)

### DIFF
--- a/src/components/DataTable/examples/07.empty-with-custom-title-and-body.jsx
+++ b/src/components/DataTable/examples/07.empty-with-custom-title-and-body.jsx
@@ -15,7 +15,7 @@ export default createClass({
 		const { data } = this.state;
 
 		return (
-			<DataTable data={data} density="extended" isFullWidth>
+			<DataTable data={data} density="extended" isFullWidth minRows={15}>
 				<EmptyStateWrapper>
 					<Title>
 						Something went wrong.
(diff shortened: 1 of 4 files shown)


chrisdopuch:
@ogupte @jondlm @sodiumjoe I could use some guidance on if I'm doing this as a non breaking change properly
My thinking was if we change the `minRows` default from 0 to 10 it causes the existing case to not break where consumers not passing `minRows` still see 10 empty rows when `data` is empty.
However, this also means that consumers who didn't pass in `minRows` before will now have 10 empty rows by default.
^^ not sure if that's really a breaking change or not?

src/components/DataTable/examples/07.empty-with-custom-title-and-body.jsx
```diff
@@ -15,7 +15,7 @@ export default createClass({
 		const { data } = this.state;
 
 		return (
-			<DataTable data={data} density="extended" isFullWidth>
+			<DataTable data={data} density="extended" isFullWidth minRows={15}>
```
ogupte: Why have this example show a custom `minRows` when there's another example that also has custom `minRows`?

chrisdopuch: That one has an incomplete page of data, so you can see minRows fill in the rest
The bug in this case was when data = [], it wouldn't fill out the minRows count, but just did a hardcoded 10 rows
so this covers that case specifically where you have no data and also minRows > 10

PR #819, opened 2017-08-23
799, add light header option to table

rsadres:
## PR Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] Edge (Win 10)


sodiumjoe:
UX wants this to be the default, but since that would be a breaking change we put it behind a top-level flag on `Table`. Next major I think we should reverse it so the dark header option is behind a flag.

chrisdopuch:
@rsadres looks like the Travis CI build failed, you'll want to look into that before merging
https://travis-ci.org/appnexus/lucid/builds/268072188?utm_source=github_status&utm_medium=notification
I think it just looks like some out of date jest snapshots

src/components/Table/Table.less
```diff
@@ -49,13 +49,55 @@
 	}
 
 	// Table header styles
+	&.lucid-Table-has-light-header > &-Thead {
```
chrisdopuch: it seems like there's a lot of repetition of styles not strictly related to the light-header. Possible to collapse this with the non-light-header table styles?

rsadres: I don't know :) I'll look at it together with one of you

PR #820, opened 2017-08-29
Aliases all of the createClass imports to createReactClass so that th...

mute:
This PR aliases all of the createClass imports to createReactClass so that they work with open source babel tooling.

The specific plugin I want consuming applications to be able to leverage is this one: https://github.com/oliviertassinari/babel-plugin-transform-react-remove-prop-types

With that said, though, it seems likely that there may be other transform tools in the future that make assumptions based on the documented conventions -- React docs suggest calling it `createReactClass` here: https://facebook.github.io/react/docs/react-without-es6.html

The ideal state would actually be if we could switch to using ES6 classes but I think that's a little more controversial.

## PR Checklist

  - [x] Chrome


mute:
Closing -- we depend on `propTypes` at runtime for `omitProps`.

PR #824, opened 2017-09-06
Sidebar: fix gripper icon centering

### LINKED ISSUE #801: Add a pixel to Sidebar gripper
Add a pixel to the Sidebar gripper area so that the gripper icon can be centered within the space. Currently the gripper icon is more toward the right of the space.

sodiumjoe:
fixes #801 

http://docspot.devnxs.net/projects/lucid/801-sidebar-gripper/

## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [ ] Edge (Win 10)

### DIFF
--- a/src/components/Sidebar/Sidebar.jsx
+++ b/src/components/Sidebar/Sidebar.jsx
@@ -294,10 +294,7 @@ const Sidebar = createClass({
 						{barProps.children}
 					</div>
 				</BarPane>
-				<SplitVertical.Divider
-					className={cx('&-Divider')}
-					style={{ width: '7px' }}
-				>
+				<SplitVertical.Divider className={cx('&-Divider')} style={{ width: 7 }}>
 					<GripperVerticalIcon className={cx('&-Divider-gripper')} />
 				</SplitVertical.Divider>
 				<PrimaryPane
--- a/src/components/Sidebar/Sidebar.jsx
+++ b/src/components/Sidebar/Sidebar.jsx
@@ -294,7 +294,10 @@ const Sidebar = createClass({
 						{barProps.children}
 					</div>
 				</BarPane>
-				<SplitVertical.Divider className={cx('&-Divider')}>
+				<SplitVertical.Divider
+					className={cx('&-Divider')}
+					style={{ width: '7px' }}
+				>
 					<GripperVerticalIcon className={cx('&-Divider-gripper')} />
 				</SplitVertical.Divider>
 				<PrimaryPane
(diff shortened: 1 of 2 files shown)


src/components/Sidebar/Sidebar.jsx
```diff
@@ -294,7 +294,10 @@ const Sidebar = createClass({
 						{barProps.children}
 					</div>
 				</BarPane>
-				<SplitVertical.Divider className={cx('&-Divider')}>
+				<SplitVertical.Divider
+					className={cx('&-Divider')}
+					style={{ width: '7px' }}
```
ogupte: `style={{ width: 7 }}`

also, why not set this in Sidebar.less?

sodiumjoe: because it's in inline style in `SplitVertical.Divider`

sodiumjoe: fixed

src/components/Sidebar/Sidebar.jsx
```diff
@@ -294,7 +294,7 @@ const Sidebar = createClass({
 						{barProps.children}
 					</div>
 				</BarPane>
-				<SplitVertical.Divider className={cx('&-Divider')}>
+				<SplitVertical.Divider className={cx('&-Divider')} style={{ width: 7 }}>
```
ogupte: why does this need to be hardcoded in the render function rather than defined in Sidebar.less?

sodiumjoe: because in `SplitVertical`:

```
										style={{
											width: '6px',
											boxSizing: 'border-box',
											...dividerProps.style,
										}}
```

chrisdopuch: The problem is that the jsx styles in SplitVertical would overwrite styles from the Sidebar.less?

sodiumjoe: that's correct

ogupte: The intent of SplitVertical is to style the divider with a default width to be overwritten by a consumer like Sidebar...You could increase the specificity of the selector in less, but it's probably best to descrease the specificity in SplitVertical.less

In any case, this should work for now, but it should be fixed with better css patterns.

sodiumjoe: it's not a selector in less, it's an inline style in jsx. The only way to override it is to pass it in as a style.

PR #834, opened 2017-10-03
831: Add navigation and variable width styling props for Tabs

astropunk:
## PR Checklist

  - [/] Chrome
  - [/] Firefox
  - [ ] Safari
  - [ ] Edge (Win 10)
- [/ ] PR has one of the `semver-` labels

### DIFF
--- a/src/components/Tabs/Tabs.jsx
+++ b/src/components/Tabs/Tabs.jsx
@@ -139,7 +139,7 @@ const Tab = createClass({
 								fill="none"
 								stroke="#fff"
 								strokeWidth="1"
-								points={isNavigation ? '0,37 7.9,18.5 0,0' : '0,28 7.9,14 0,0'}
+								points={isNavigation ? '0,37 7.3,18.5 0,0' : '0,28 7.9,14 0,0'}
 							/>
 						</svg>
 					</span>}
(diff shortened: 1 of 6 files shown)


astropunk:
screen shot 2017-10-03 at 2 28 44 pm

navigation + variable width + progression: 
screen shot 2017-10-03 at 2 14 07 pm

rsadres:
Can we make it so that the separator icon is touching the top and bottom borders please, like in the older Tabs.
Thank you

rsadres:
We need to create a new arrow svg for this new tab height. Working on it.

src/components/Tabs/Tabs.jsx
```diff
@@ -71,6 +71,16 @@ const Tab = createClass({
 		 * The content to be rendered as the `Title` of the `Tab`.
 		 */
 		Title: node,
+
+		/**
+		 * Width (with unit) to be passed as the `style` to the `Tab`
+		 */
+		width: string,
```
ogupte: Instead of having a specific prop for width, passing in a `style` prop with a `width` property seems more idiomatic. We do use specific dimension props like `width` in cases where the render function does some logic based on that width value. In this case, it's being pass thru directly to the `style` prop anyways, so might as well keep that pattern.

astropunk: I can do that. What I was going for with the `width` prop was to create an interface that declares width as an intended style-override and use-case; where a style prop is a general escape hatch and users may or may not know which styles are intended to be overridden.

PR #843, opened 2017-10-27
Perf improvements for DateSelect and ConextMenu

ogupte:
- Adds new prop `useSlidePanel` to DateSelect for finer control of performance
 - reduces number of re-renders and state changes in ContextMenu
 - ContextMenu FlyOut now uses requestAnimationFrame to synchronize alignments

## PR Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] Edge (Win 10)

### DIFF
--- a/src/components/Autocomplete/Autocomplete.spec.jsx
+++ b/src/components/Autocomplete/Autocomplete.spec.jsx
@@ -10,6 +10,17 @@ import { DropMenuDumb as DropMenu } from '../DropMenu/DropMenu';
 import * as KEYCODE from '../../constants/key-code';
 
 describe('Autocomplete', () => {
+	let requestAnimationFramePreviewValue;
+
+	beforeAll(() => {
+		requestAnimationFramePreviewValue = window.requestAnimationFrame;
+		window.requestAnimationFrame = callback => setTimeout(callback, 32);
+	});
+
+	afterAll(() => {
+		window.requestAnimationFrame = requestAnimationFramePreviewValue;
+	});
+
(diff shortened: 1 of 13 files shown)


src/components/Autocomplete/Autocomplete.spec.jsx
```diff
@@ -10,6 +10,17 @@ import { DropMenuDumb as DropMenu } from '../DropMenu/DropMenu';
 import * as KEYCODE from '../../constants/key-code';
 
 describe('Autocomplete', () => {
+	let requestAnimationFramePreviewValue;
+
+	beforeAll(() => {
+		requestAnimationFramePreviewValue = window.requestAnimationFrame;
+		window.requestAnimationFrame = callback => setTimeout(callback, 32);
```
sodiumjoe: for the react 16 stuff, I'm adding this to an init script that runs before all the tests, should probably just do that here

src/components/DateSelect/DateSelect.jsx
```diff
@@ -164,6 +171,7 @@ const DateSelect = createClass({
 			onSelectDate: _.noop,
 			isFontSizeRelative: false,
 			showCursorHighlight: true,
+			useSlidePanel: true,
```
chrisdopuch: Do we want to make the default value be the more performant one?

ogupte: it would be considered a breaking change if we default it to false since it disables touch support.

Issue #797, opened 2017-07-12
Update RadioButtonLabeled and CheckboxLabeled clickable area


rkaneriya:
Update components so that the radio button and it's label are the click target for RadioButtonLabeled and CheckboxLabeled.

I've noticed a couple tickets crop up in the past about clients getting confused by our inconsistency re: "clickable areas" on radio buttons. Sometimes, the clickable area is restricted only to the button (or checkbox) itself, sometimes it's the button + its label, and sometimes it's the button + its label + everything to the right. That's a problem our apps should address. 

To help in that consistency effort, I suggest we reduce the clickable area on RadioButtonLabeled and CheckboxLabeled components to only the button/checkbox + its label. I believe that is the UX community's consensus on this. Ex: https://uxplanet.org/radio-buttons-ux-design-588e5c0a50dc.

rsadres:
![screenshot 2017-11-15 12 09 45](https://user-images.githubusercontent.com/10146234/32849689-e86f0614-c9fd-11e7-8090-967dcedf0ab3.png)

PR #868, opened 2017-11-30
RFC: #797 Reduce clickable area of CheckboxLabeled

ogupte:
Don't merge, this PR is for just for feedback.

A simple solution is to wrap the existing <label> element so that the label doesn't get stretched to the width of the container. Thoughts?

### DIFF
--- a/src/components/CheckboxLabeled/CheckboxLabeled.jsx
+++ b/src/components/CheckboxLabeled/CheckboxLabeled.jsx
@@ -80,30 +80,33 @@ const CheckboxLabeled = createClass({
 		);
 
 		return (
-			<label
-				className={cx(
-					'&',
-					{
-						'&-is-disabled': isDisabled,
-						'&-is-selected': isIndeterminate || isSelected,
-					},
-					className
-				)}
-				style={style}
-			>
-				<Checkbox
-					className={cx('&-Checkbox', className)}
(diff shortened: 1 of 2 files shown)


sodiumjoe:
if we wanted to go this route, why couldn't we just put `display: flex` on the label.

ogupte:
>if we wanted to go this route, why couldn't we just put display: flex on the label.

That is in fact [how it exists today](https://github.com/appnexus/lucid/blob/master/src/components/CheckboxLabeled/CheckboxLabeled.less#L8)

But it results in the `<label>` being a block element with a full-width click target. My wrapper approach here makes the `<label>` element width the same as the content, which minimizes the click area, but since the wrapper is a block element, it retains its block-level, line-breaking behavior that would otherwise be lost by setting `display: inline-flex` on the `<label>`.

I think there's a different discussion to be had about whether these elements need to be block elements vs inline elements, but the wrapper ensures that it won't break existing layouts.

sodiumjoe:
I see, I guess that's why it makes most sense to me for consuming applications to add the wrapper, as the grouping element. But I'm ok with this too.

ogupte:
>...I guess that's why it makes most sense to me for consuming applications to add the wrapper, as the grouping element...

I agree, but it would have to be something we address in the next major version of the library.

PR #873, opened 2017-12-14
851 icon group

alisonailea:
## PR Checklist

- Manually tested across supported browsers
  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [ ] Edge (Win 10)
- [x] Unit tests written (`common` at minimum)
- [x] PR has one of the `semver-` labels
- [ ] Two core team engineer approvals
- [ ] One core team UX approval

PR #874, opened 2017-12-22
IconGroup Feedback and notes

ogupte:
This is feedback for PR #873

I added notes for things that I thought about after spending some time trying trying to simplify the component API.

I also added 3 examples of a suggested component API that handles the 3 main use cases: selecting one icon, selecting many icons, partial selection of icons.

### DIFF
--- a/high-level-feedback.md
+++ b/high-level-feedback.md
@@ -1,0 +1,27 @@
+# Recategorize
+	- Is used more like a selector (like SingleSelect & MultiSelect) than a toggle
+	- update name to something like IconSelect since it handles both single or multi select (using radio or checkboxes)
+# Keep it stateless
+	The minimal application state varies widely between various use cases:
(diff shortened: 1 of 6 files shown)


high-level-feedback.md
```diff
@@ -0,0 +1,27 @@
...
+	Leave partial selection behavior up to the consumer
+		- items can be rendered with `isPartial` prop, `onSelect` callback passes name only
+			onSelect: (state, name) => {
+				const isPartial = findIcon(name).isPartial;
+				if (isPartial) { ... }
+				return nextState;
+			}
+# One component
+	- IconGroup is basically a wrapper for state management and container element
+	- IconBox is the real star component, but it doesn't have to be singular
```
ogupte: you could alternatively break it out into more than one component, but only expose one interface

PR #883, opened 2018-01-20
Generate docs by referencing components

ogupte:
## PR Checklist

  - [x] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] Edge (Win 10)


package.json
```diff
@@ -10,12 +10,13 @@
   "scripts": {
     "build-bundle": "scripts/build-bundle.sh",
     "cover": "npm run jest -- --coverage",
-    "dev": "gulp docs-generate && webpack-dev-server --inline --hot --content-base src/docs/",
+    "dev": "webpack-dev-server --inline --hot --content-base src/docs/",
```
sodiumjoe: is this useful? why not just make the `peek` task `dev` instead?

src/docs/index.jsx
```diff
@@ -137,9 +118,9 @@ const PropType = createClass({
 		const { componentName, type, shapeKey } = this.props;
 
 		if (!type) {
-			console.error(
-				`It looks like there's an issue with ${componentName}'s props. One common cause of this issue is when the getDefaultProps method defines a default value for a prop that is not explicitly defined in the propTypes map.`
-			);
+			//console.error(
```
sodiumjoe: probz just `delet`?

src/docs/index.jsx
```diff
@@ -187,6 +169,38 @@ const Component = createClass({
...
+				componentRef={lucidComponents[componentName]}
+				location={location}
+			/>
+		);
+	},
+});
+
+const isReactComponent = value => {
+	try {
+		return (
```
sodiumjoe: what would `throw` here?

src/docs/index.jsx
```diff
@@ -210,58 +224,39 @@ const Component = createClass({
...
-				);
-			}
-		);
+		const { componentName, componentRef } = this.props;
+		const childComponents = _.toPairs(_.pickBy(componentRef, isReactComponent));
+		const defaultProps = componentRef.defaultProps;
 
 		const componentProps = _.flow(
-			x => _.get(x, `${componentName}.props`, []),
+			x => _.get(x, `propTypes`, {}),
```
sodiumjoe: why template literal

src/docs/index.jsx
```diff
@@ -210,58 +224,39 @@ const Component = createClass({
...
-			`${componentName}.description`
-		);
+		const descriptionMarkdown = _.get(componentRef, 'peek.description', '');
 
-		if (!descriptionMarkdown) {
-			throw new Error(
-				`Unable to find ${componentName}'s description in src/docs/docgen.json. Are you on a bad URL? Did you forget to run \`gulp docs-generate\`? Did you remember to give ${componentName} a description block?`
-			);
-		}
+		//if (_.isEmpty(descriptionMarkdown)) {
```
sodiumjoe: `delet`

src/docs/index.jsx
```diff
@@ -299,36 +296,35 @@ const Component = createClass({
 						</Thead>
 						<Tbody>
 							{_.map(componentProps, ([propName, propDetails]) => {
-								if (!propDetails || !propDetails.description) {
-									console.error(
-										`Warning: There was an issue with the docs that were generated for component "${componentName}" and prop "${propName}". One reason might be that you have a default value for something that was never declared in propTypes.`
-									);
-									return null;
-								}
+								//if (!propDetails.text) {
```
sodiumjoe: `delet`

package.json
```diff
@@ -49,11 +50,11 @@
     "prop-types": "^15.5.10",
```
sodiumjoe: can drop this one now, right?

package.json
ogupte: whoops i didn't mean to leave that in there, that was for my own local testing only. will remove

src/docs/index.jsx
ogupte: whoops, the try/catch is old code from an earlier attempt to check if the value was a component. will remove.

src/docs/index.jsx
```diff
@@ -561,51 +550,57 @@ const App = createClass({
...
-
-			if (componentName.toLowerCase().indexOf(search.toLowerCase()) !== -1) {
-				return componentName;
+		const searchLower = _.toLower(search);
+		const showPrivateComponents = this.showPrivateComponents();
+
+		return _.filter(_.keys(lucidComponents), key => {
+			const componentRef = lucidComponents[key];
+			const isPrivate =
+				_.get(componentRef, 'peek.isPrivate') ||
```
chrisdopuch: How does peek map to private? would private be set on a prop?

src/docs/index.jsx
```diff
@@ -561,51 +550,57 @@ const App = createClass({
...
-				// exclude child components from grouping
-				if (_.includes(key, '.')) {
-					return acc;
+		const categoryTree = _.reduce(
+			lucidComponents,
+			(acc, value, key) => {
+				const categories = _.get(value, 'peek.categories', []);
+				const path = categories.concat(key).join('.');
+				const isPrivate =
+					_.get(value, 'peek.isPrivate') || _.get(value, '_isPrivate');
```
chrisdopuch: could pull this out into a selector function, since we use it a couple places

src/docs/index.jsx
```diff
@@ -561,51 +550,57 @@ const App = createClass({
...
+					style={{
+						position: 'fixed',
+						left: 0,
+						top: 0,
+						width: '100%',
+						height: '100%',
+						backgroundColor: 'white',
+						zIndex: 2,
+						pointerEvents: 'none',
+						transition: 'opacity 200ms ease 200ms',
```
chrisdopuch: Neat!

chrisdopuch: Any reason not to just put this in index.less since we're already styling it there too?

ogupte: The reason the FOUC happens in the first place is because the page renders before the styles are loaded.

src/docs/index.jsx
ogupte: the `isPrivate` flag is set on proptypes resolvers which are accessed from the `.isRequired` reference of the normal resolver.

PR #892, opened 2018-02-05
891: add tooltip formatter to bars and bar charts

ToddKranenburg:
## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [ ] Edge (Win 10)


src/components/Bars/Bars.jsx
```diff
@@ -267,26 +276,30 @@ const Bars = createClass({
...
-												>
-													{yTooltipFormatter(
-														_.get(legend, field, field),
-														yFormatter(data[seriesIndex][field]),
-														data[seriesIndex][field]
-													)}
-												</Legend.Item>
-											))}
-										</Legend>
+										{tooltipFormatter
```
sodiumjoe: I think this would be a little cleaner if you pulled the default toolTipFormatter up into a `this.defaultToolTipFormatter`, and used that as the default value for `this.props.toolTipFormatter`. What do you think?

ToddKranenburg: Makes sense to me!

ToddKranenburg: Just gave this refactor a go - `getDefaultProps` doesn't have access to instance variables so can't reference `this.defaultTooltipFormatter` there. I could either define the default formatter within the `getDefaultProps` scope and add a `props` argument (since we don't have access to `this.props` in that scope) or leave it as is... maybe with a small refactor like: ```{tooltipFormatter ? tooltipFormatter(data[seriesIndex]) : defaultTooltipFormatter(data[seriesIndex])}```

Will defer to your preference on this! Or maybe you see a cleaner way to deal with this that I don't.

ogupte:
`toolTipFormatter` is essentially a render function as a prop, so it might be good to be clear about that by calling it `renderToolTipBody` or something. That way it's understood that it doesn't just compete with other formatter functions (which expect strings to be returned) but that it entirely replaces all the content in a tool tip body.

PR #895, opened 2018-02-09
Publish docgen json

ogupte:
Add gulp task to generate docgen json and publish it in dist

### DIFF
--- a/gulp/docs.js
+++ b/gulp/docs.js
@@ -1,5 +1,6 @@
 /*eslint no-console:0 */
 
+var _ = require('lodash');
 var FormData = require('form-data');
 var exec = require('child_process').exec;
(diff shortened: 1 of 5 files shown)


gulp/docs.js
```diff
@@ -9,12 +10,112 @@ var path = require('path');
 var tar = require('gulp-tar');
 var webpack = require('webpack');
 var webpackConfig = require('../webpack.config.js');
+var CONFIG = require('./config');
+
+const isReactComponent = value =>
+	typeof value === 'function' &&
```
sodiumjoe: are the indents weird here?

ogupte: i agree, but this is what prettier wants :P

PR #899, opened 2018-02-16
optimize performance of bar charts

sodiumjoe:
## PR Checklist

  - [x] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] Edge (Win 10)

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,6 +47,7 @@
     "element-resize-detector": "^1.1.6",
     "lodash": "^4.11.0",
     "prop-types": "^15.6.0",
+    "react-addons-shallow-compare": "^15.6.2",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -47,6 +47,7 @@
     "element-resize-detector": "^1.1.6",
     "lodash": "^4.11.0",
     "prop-types": "^15.6.0",
+    "react-addons-shallow-compare": "^15.6.2",
```
ogupte: nice!

sodiumjoe: it's... actually kinda not nice, because it doesn't look like the package is maintained. But it's also temporary, once we move to react class syntax, we can just use `extend React.PureComponent` or whatever

PR #918, opened 2018-04-06
Build upgrade (Webpack 4 & more)

ogupte:
upgrade bundle build to webpack 4 & loaders and refactor build logic out of gulp and into webpack + scripts


scripts/docs-upload-gh.sh
```diff
@@ -1,36 +0,0 @@
-#!/bin/bash
```
sodiumjoe: are we not uploading to github pages anymore?

ogupte: we are, i just moved the logic to `scripts/upload-docs-build-gh.sh` to be consistent with the docspot upload script

scripts/path-import-modules.js
```diff
@@ -0,0 +1,167 @@
...
+				});
+			});
+		})
+	);
+
+	return checkFilesPromise.then(writeFilesPromise);
+};
+module.exports.generateModules = generateModules;
+
+const removeModules = (entryModulePath, rootPath = '.') => {
```
chrisdopuch: Could you accomplish this with a simple `git clean -fd` or something like that instead?

scripts/upload-docs-build-docspot.sh
```diff
@@ -0,0 +1,37 @@
...
+popd &>/dev/null
+
+echo "POSTing to docspot (buildId: $BUILD_ID, isLatest: $IS_LASTEST)..."
+curl "http://docspot.devnxs.net/api/projects" \
+	--form file=@/tmp/$BUILD_ID.tar.gz \
+	--form projectId=anx-react \
+	--form buildId=$BUILD_ID \
+	--form isLatest=$IS_LASTEST &>/dev/null
+
+if [ $? -eq 0 ]; then
```
chrisdopuch: What happens in case of failure here? Does it just swallow the error?

scripts/path-import-modules.js
ogupte: possibly, however i primed this script to be externalized in the near future, so that we can leverage it in other projects, and we don't want to make the assumption that all non-staged files should be removed.

scripts/upload-docs-build-docspot.sh
ogupte: good point, we should add messaging for that case

webpack.config.bundle.js
```diff
@@ -1,61 +1,50 @@
...
+					{ loader: 'less-loader', options: { sourceMap: !isProduction } },
+				],
 			},
 		],
 	},
 	resolve: {
-		extensions: ['', '.js', '.jsx'],
+		extensions: ['.js', '.jsx'],
+	},
+	devtool: isProduction ? false : 'source-map',
```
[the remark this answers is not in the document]
ogupte: we could add them, but it's pretty standard to only include source maps for the development build.

PR #922, opened 2018-04-24
Selector option formatting

ogupte:
Added more formatting options to `SingleSelect`, `SearchableSelect`, and `SearchableMultiSelect`:
- Added new component `Underline` that will underline child text that matches a given text pattern
- `Option` accepts a `Selected` prop/child that is rendered when the option is selected
- `Option` children accepts a function/component as children which is passed the current `searchText` as a prop

## PR Checklist

  - [x] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] Edge (Win 10)

### DIFF
--- a/src/components/DropMenu/DropMenu.jsx
+++ b/src/components/DropMenu/DropMenu.jsx
@@ -283,7 +283,7 @@ const DropMenu = createClass({
 			statics: {
 				peek: {
 					description: `
-						Props that are passed through to the underling ContextMenu.
+						Props that are passed through to the underlying ContextMenu.
 					`,
 				},
 			},
--- a/src/components/SearchableMultiSelect/SearchableMultiSelect.jsx
+++ b/src/components/SearchableMultiSelect/SearchableMultiSelect.jsx
@@ -67,7 +67,14 @@ const SearchableMultiSelect = createClass({
 				},
 			},
 			propName: 'Option',
-			propTypes: DropMenu.Option.propTypes,
+			propTypes: {
+				filterText: string`
+					Text used to filter options when searching. By default, this is the
+					text rendered in the Option, but it can be customized further with
+					this prop.
+				`,
+				...DropMenu.Option.propTypes,
+			},
 			components: {
 				Selected: createClass({
 					displayName: 'SearchableMultiSelect.Option.Selected',
--- a/src/components/SearchableMultiSelect/SearchableMultiSelect.jsx
+++ b/src/components/SearchableMultiSelect/SearchableMultiSelect.jsx
@@ -69,6 +69,18 @@ const SearchableMultiSelect = createClass({
 			propName: 'Option',
 			propTypes: DropMenu.Option.propTypes,
 			components: {
+				Selected: createClass({
+					displayName: 'SearchableMultiSelect.Option.Selected',
+					statics: {
+						peek: {
+							description: `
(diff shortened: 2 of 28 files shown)


src/components/DropMenu/DropMenu.jsx
```diff
@@ -283,7 +283,7 @@ const DropMenu = createClass({
 			statics: {
 				peek: {
 					description: `
-						Props that are passed through to the underling ContextMenu.
```
chrisdopuch: haha ok but I kind of liked it before

src/components/Underline/Underline.jsx
```diff
@@ -0,0 +1,68 @@
...
+
+const cx = lucidClassNames.bind('&-Underline');
+
+const { node, string, instanceOf, oneOfType } = PropTypes;
+
+const matchAllRegexp = /^.*$/;
+
+const Underline = ({ className, children, match }) => {
+	if (!_.isRegExp(match)) {
+		if (_.isString(match)) {
```
chrisdopuch: this looks like a good candidate for a ternary
also, is it a good idea to mutate `match` instead of creating a local var?

ogupte: it doesn't mutate the actual match that is passed in, it just reassigns a new value to that reference. i could create a local var instead of reusing the reference if that makes it more clear.

src/components/Underline/Underline.jsx
```diff
@@ -0,0 +1,68 @@
...
+			match = matchAllRegexp;
+		}
+	}
+
+	if (!_.isString(children)) {
+		return (
+			<span className={cx('&', className)}>
+				<span
+					style={
+						match === matchAllRegexp ? { textDecoration: 'underline' } : null
```
sodiumjoe: nit: should pull these style object literals up into a module constant

src/components/SearchableMultiSelect/SearchableMultiSelect.jsx
```diff
@@ -380,7 +406,9 @@ const SearchableMultiSelect = createClass({
 					<CheckboxLabeled.Label>
 						{_.isString(optionProps.children)
 							? this.renderUnderlinedChildren(optionProps.children, searchText)
```
sodiumjoe: can't we replace these with `<Underline />`?

src/components/SearchableMultiSelect/SearchableMultiSelect.jsx
```diff
@@ -100,6 +112,20 @@ const SearchableMultiSelect = createClass({
 			},
 			propName: 'OptionGroup',
 			propTypes: DropMenu.OptionGroup.propTypes,
+			components: {
+				Selected: createClass({
```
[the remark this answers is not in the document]
ogupte: It is in the sense that it shows the OptionGroup along with other SelectedItems below the drop down

src/components/SearchableMultiSelect/SearchableMultiSelect.jsx
ogupte: i'd like to, but it does break snapshots since the shallow rendering + class names are changed

sodiumjoe: 🤔 yeah, not sure if we should consider that a breaking change

PR #923, opened 2018-04-24
Anxr1554 optimization

sodiumjoe:
## PR Checklist

  - ~[ ] Chrome~
  - ~[ ] Firefox~
  - ~[ ] Safari~
  - ~[ ] Edge (Win 10)~
- ~[ ] Unit tests written (`common` at minimum)~
- ~[ ] One core team UX approval~


chrisdopuch:
So the idea is that we specify just the targets we desire for transpilation, so that the final bundle is smaller due to fewer polyfills?

sodiumjoe:
@chrisdopuch that's the preset-env part, yeah. The other part is a separate build that transpiles everything except es-modules syntax. This allows downstream webpack tree-shaking to propagate through lucid.

PR #927, opened 2018-05-03
add title to checkbox

jhsu:
use `title` on the wrapper element so that a tooltip can be shown on hover.

### DIFF
--- a/src/components/Checkbox/Checkbox.jsx
+++ b/src/components/Checkbox/Checkbox.jsx
@@ -56,6 +56,10 @@ const Checkbox = createClass({
 		style: object`
 			Passed through to the root element.
 		`,
+
+		title: string`
+			A string title that is displayed on hover.
+		`,
 	},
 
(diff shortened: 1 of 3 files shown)


src/components/Checkbox/Checkbox.jsx
```diff
@@ -85,6 +89,8 @@ const Checkbox = createClass({
 			...passThroughs
 		} = this.props;
 
+		const { title } = passThroughs;
```
sodiumjoe: why not just add to the destructure statement above?

jhsu: because we still want to pass it to the input

sodiumjoe: i think it still gets omitted by `omitProps`, since it's declared in proptypes.

jhsu: you're right, let me look into it

PR #930, opened 2018-05-21
Portal: pass store context to children

ogupte:
In Portal, pass along the `store` context property to rendered Portal children

## PR Checklist

  - [x] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] Edge (Win 10)


ogupte:
> Could this have any adverse affects on users who aren't using a context?
if there's no context, the only difference is an additional nested component level, but not an additional DOM element level

PR #931, opened 2018-05-22
Support for npm link with path imports and style dependencies

ogupte:


package.json
```diff
@@ -19,6 +19,8 @@
     "dev": "npm run storybook",
     "docs": "NODE_ENV=documentation start-storybook -s ./docs/public -p 6006",
     "jest": "NODE_ENV=test TZ=UTC jest --runInBand",
+    "link": "node ./scripts/path-import-source-modules-exec.js --generate",
```
chrisdopuch: These new commands would be useful to document in the readme

jondlm:
Need to vet these changes in a consuming application. I'm a little concerned that the changes to the less imports could add a ton of duplication into the final CSS.

ogupte:
The NPM linking and style imports are very intertwined in this PR. I'm going to close this PR due to my inactivity, but the branch `anxr814-npm-link` will remain if we wanted to continue down this path for handling the styles. By now, the core team probably have a better approach for handling linking during app development.

PR #932, opened 2018-05-23
fixes #523 by changing the styles

### LINKED ISSUE #523: Add hasGutters to Dialog
Add a new prop to the `Dialog` component. It should have a default value of `true` so we don't break existing usage of the component. The new prop should be called `hasGutters` and it should take a boolean. If set to `false` it should remove the padding from the content inside the `Dialog` as seen in the image below:

image

liliedelamotte:
## PR Checklist

  - [x] Chrome
  - [ ] Firefox
  - [x] Safari
  - [ ] Edge (Win 10)

### DIFF
--- a/src/components/Dialog/Dialog.jsx
+++ b/src/components/Dialog/Dialog.jsx
@@ -7,11 +7,13 @@ import { createClass, getFirst, omitProps } from '../../util/component-types';
 
 const cx = lucidClassNames.bind('&-Dialog');
 
-const { node, oneOf } = PropTypes;
+const { node, oneOf, bool } = PropTypes;
 
 const SMALL = 'small';
 const MEDIUM = 'medium';
 const LARGE = 'large';
+// hasGutters added to allow for a choice regarding Body padding
+var hasGutters;
 
 const Dialog = createClass({
 	displayName: 'Dialog',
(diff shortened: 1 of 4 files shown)


src/components/Dialog/Dialog.less
```diff
@@ -53,6 +53,13 @@
 		max-height: 60vh; // fix for IE 11 bug, otherwise the body doesn't shrink properly
 	}
 
+	&-body-no-gutters {
+		flex: 1 1 auto;
```
jondlm: If you keep the `&-body` class and conditionally add the `&-body-no-gutters` class you should be able to make this block just have `padding: 0` I think.

src/components/Dialog/Dialog.jsx
```diff
@@ -70,16 +72,27 @@ const Dialog = createClass({
...
 	},
 
 	getDefaultProps() {
 		return {
 			size: MEDIUM,
+			hasGutters: true,
 		};
 	},
 
 	render() {
```
joelgriffith: Nice work on the destructuring + `...passThroughs`

src/components/Dialog/Dialog.less
jhsu: or you'll probably want `&-body.&-body-no-gutters {`

PR #934, opened 2018-05-31
fixes #796 by updating Banner remove/close icon

### LINKED ISSUE #796: Update banner remove/close icon
Modify the close X on [Banner](https://appnexus.github.io/lucid/?selectedKind=Banner&selectedStory=stateless&full=0&addons=1&stories=1&panelRight=1&addonPanel=lucid-docs-panel-props) to be the same as the close X on [Selection](https://appnexus.github.io/lucid/?selectedKind=Selection&selectedStory=kinds&full=0&addons=1&stories=1&panelRight=1&addonPanel=lucid-docs-panel-props)

## Open questions

- Do we need to change the color depending on the banner background color? Ping @rsadres if it doesn't look right.

liliedelamotte:
## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [ ] Edge (Win 10)

### DIFF
--- a/src/components/Banner/Banner.less
+++ b/src/components/Banner/Banner.less
@@ -9,7 +9,7 @@
 @Banner-size-iconFontSize: 16px;
 
 .lucid-Banner {
-	.transition-group-animation-fade();
+	// .transition-group-animation-fade();
 	.box-sizing();
 
 	border: @border-standardBorder;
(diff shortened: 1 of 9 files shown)


src/components/Banner/Banner.less
```diff
@@ -28,17 +28,12 @@
 		border-radius: 0;
 	}
```
joelgriffith: I didn't have good idea if there's a preferred convention for overriding prior-set CSS styles, so we went with "beefing" up this selector so it overrode the prior

jondlm:
Looks like there are some snapshot tests that need to updated.

src/components/Banner/Banner.less
jondlm: Although this works, generally we try to add a new class to the existing component and reference it that way. In other words, you should be able to replace `.lucid-Icon-is-clickable` with `.lucid-Banner-close` since you've already added that class.

This prevents the case where someone putting a different icon inside their own banner might have their styles stomped on.

PR #937, opened 2018-06-18
fixes #741 by adding onSelect handler to Icon

### LINKED ISSUE #741: Add onSelect to Icon
It'd be nice to have an `onSelect` handler for `Icon`s with the `({ event, props}) => {}` signature for cases where we need to pluck off `callbackId` from them. Typically we'd just do `onClick` but given that icons have been out in the wild for a while, I think it would be potentially damaging to overwrite the `onClick` handler and change the signature.

liliedelamotte:
## PR Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] Edge (Win 10)

### DIFF
--- a/src/components/Icon/Icon.jsx
+++ b/src/components/Icon/Icon.jsx
@@ -1,12 +1,13 @@
 import _ from 'lodash';
 import React from 'react';
 import PropTypes from 'react-peek/prop-types';
+import ReactDOM from 'react-dom';
 import { lucidClassNames } from '../../util/style-helpers';
 import { createClass, omitProps } from '../../util/component-types';
 
 const cx = lucidClassNames.bind('&-Icon');
 
-const { any, string, number, object, bool } = PropTypes;
+const { any, string, number, object, bool, func } = PropTypes;
 
 const Icon = createClass({
 	displayName: 'Icon',
@@ -60,6 +61,11 @@ const Icon = createClass({
 			isClickable to be false.
 		`,
 
+		onClick: func`
+			Called when user clicks the \`Icon\`. Signature:
+			\`({event, props}) => {}\`
+		`,
+
 		children: any`
 			Any valid React children.
 		`,
@@ -72,7 +78,18 @@ const Icon = createClass({
 			viewBox: '0 0 16 16',
 			isBadge: false,
 			isDisabled: false,
+			isClickable: false,
 		};
+	},
+
+	onSelect(event) {
+		const { isClickable, isDisabled, onClick } = this.props;
+		const domNode = ReactDOM.findDOMNode(this);
+
+		if (!isDisabled && isClickable) {
+			domNode.focus();
+			onClick({ event, props: this.props });
+		}
 	},
 
 	render() {
@@ -118,6 +135,7 @@ const Icon = createClass({
 					},
 					className
 				)}
+				onClick={this.handleSelect}
 			>
 				{children}
 			</svg>
--- a/src/components/Icon/Icon.jsx
+++ b/src/components/Icon/Icon.jsx
@@ -61,7 +61,11 @@ const Icon = createClass({
 			isClickable to be false.
 		`,
 
-		onClick: func`
+		onSelect: func`
+			Called if user already has \`onClick\` in place.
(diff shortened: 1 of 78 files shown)


src/components/Icon/Icon.jsx
```diff
@@ -60,6 +61,11 @@ const Icon = createClass({
 			isClickable to be false.
 		`,
 
+		onClick: func`
```
liliedelamotte: Since onClick is a function and not a propType, does this bit need to be here?

src/components/Icon/Icon.jsx
```diff
@@ -118,6 +135,7 @@ const Icon = createClass({
 					},
 					className
 				)}
+				onClick={this.handleSelect}
```
liliedelamotte: Is `onClick` the correct term to use here? If the goal is to create an `onClick` handler but to also prevent overriding any other handlers downstream, we use the term `onSelect` for the handler but `onClick` for the actual action, correct?

src/components/Icon/Icon.jsx
jondlm: It probably wasn't super clear from my ticket description, but this should actually be `onSelect`. We want the component _users_ to start using `onSelect` rather than `onClick`. In fact, many of them are probably already using `onClick` today (because of the way we `passThrough` unknown props) and we have to be careful not to break their existing usage of `onClick`.

ogupte: callback functions are also defined as propTypes so this is correct

src/components/Icon/Icon.jsx
ogupte: `onClick` must be used here since it's on an `svg` html element which accepts `onClick` (see https://reactjs.org/docs/dom-elements.html#all-supported-html-attributes)

src/components/Icon/Icon.jsx
```diff
@@ -72,9 +78,20 @@ const Icon = createClass({
 			viewBox: '0 0 16 16',
 			isBadge: false,
 			isDisabled: false,
+			isClickable: false,
 		};
 	},
 
+	onSelect(event) {
```
ogupte: maybe rename this method to `handleClick` to be consistent with `onClick`

src/components/Icon/Icon.jsx
ogupte: this.handleSelect isn't defined, so this would cause an error. it looks like `this.onSelect` is defined above but also consider renaming that method to `this.handleClick` for consistency

src/components/Icon/Icon.jsx
```diff
@@ -72,9 +78,20 @@ const Icon = createClass({
...
+		const { isClickable, isDisabled, onClick } = this.props;
```
jondlm: You're right to pluck out `onClick` but we should have some code like this:

```javascript
if (onClick) {
  onClick(event);
}
```

That way existing users who are currently relying on `onClick` to have specific arguments won't be broken. Instead we'll be introducing a new `onSelect` that has the `{ event, props }` object.

src/components/Icon/Icon.jsx
```diff
@@ -72,9 +78,20 @@ const Icon = createClass({
...
+		const domNode = ReactDOM.findDOMNode(this);
+
+		if (!isDisabled && isClickable) {
+			domNode.focus();
+			onClick({ event, props: this.props });
```
jondlm: We should call `onSelect` here instead

src/components/Icon/Icon.jsx
jondlm: You're correct to be using `onClick` here. For clarity I'd probably make this `onClick={this.handleClick}` and rename your callback function to `handleClick` (which will call the user's `onSelect` prop)

liliedelamotte:
Looks like this PR passed the Travis CI checks but not the codecov/project tests. Do those tests need to be passed? If so, what can I do to make sure that happens?

jondlm:
Went ahead and merged it. We probably should have added some unit tests but we'll let it slide this time :)

PR #938, opened 2018-06-27
Fixes #935 by adding a ProgressBar component

### LINKED ISSUE #935: ProgressBar component
Proposing adding a ProgressBar element to the lucid component set.

# Prior art

https://getbootstrap.com/docs/4.1/components/progress/
https://material.io/design/components/progress-indicators.html#linear-progress-indicators

# Use cases

* Indicates progress in a procedure consisting of numerous discrete steps or continuous operation (such as loading, saving multiple objects in the system, progress of the user through a workflow, estimated loading time, etc.
* Indicates progress towards a non-asynchronous goal, e.g. dollar amounts vs. targets

# Implementation

* This can be easily achieved by using css flex-box and using a `width` style value on an immediate inner child which will span the given width, e.g.
```
<div className='progressBar' style={{ display: 'flex' }}>
    {/* this would occupy a third of the width of the parent, as the style property suggests */}
    <div className='progressBar-indicator', style={{ width: '33%' }}>
    </div>
</div>
```

liliedelamotte:
## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [ ] Edge (Win 10)


src/components/Icon/Icon.jsx
```diff
@@ -60,6 +61,11 @@ const Icon = createClass({
 			isClickable to be false.
 		`,
 
+		onClick: func`
```
jondlm: Uh oh, looks like your icon change made their way into this PR :/

jondlm:
Let's get some unit tests added for this component. See other components for examples.

src/components/ProgressBar/ProgressBar.jsx
```diff
@@ -0,0 +1,79 @@
...
+			Percentage ProgressBar is complete.
+		`,
+
+		children: any,
+
+		hasTitle: bool`
+			Whether or not a title is displayed above the ProgressBar.
+		`,
+
+		title: string`
```
jondlm: Let's change the `title` prop to be a `Title` prop and allow folks to pass more than just text. 

Here's an example of how we did it with `ToolTip`: https://github.com/appnexus/lucid/blob/c57a348669a15fa727865241d0e4bdee0912e28e/src/components/ToolTip/ToolTip.jsx#L118

stories/ProgressBar.stories.js
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
```
jondlm: Is this file needed? I'm not sure.

docs/examples/ProgressBar/WithDefaults.js
```diff
@@ -0,0 +1,44 @@
...
+			percentEntered: '',
+		});
+		event.preventDefault();
+	}
+
+	render() {
+		return (
+			<div>
+				<form onSubmit={this.handleSubmit}>
+					<input
```
jondlm: Instead of using `input`, let's use a `TextField` instead since that's our version of the input.

jondlm:
Looks like there are some linting failures on this PR: you can see them locally with `npm run lint`

docs/examples/ProgressBar/VariousColors.jsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import createClass from 'create-react-class';
+import { ProgressBar } from '../../../src/index.js';
+
+export default createClass({
+	getInitialState() {
```
chrisdopuch: Doesn't look like this component does anything with state, can this be removed?

docs/examples/ProgressBar/WithTitle.jsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import createClass from 'create-react-class';
+import { ProgressBar } from '../../../src/index.js';
+
+export default createClass({
+	getInitialState() {
```
chrisdopuch: Looks like this one can be removed too

src/components/ProgressBar/ProgressBar.jsx
```diff
@@ -0,0 +1,109 @@
...
+	propTypes: {
+		className: string`
+			Appended to the component-specific class names set on the root element.
+		`,
+
+		kind: oneOf(['default', 'success', 'danger', 'info', 'warning'])`
+			Applies a color style for the kind of ProgressBar.
+		`,
+
+		percentComplete: number`
```
chrisdopuch: It's probably worth validating that this value is a valid percentage (i.e. a positive number between 0 and 100)

You could do that by using `oneOf` in combination with `_.range`

Something like this should work:
```
percentageComplete: oneOf(_.range(0, 101))
```

(this does assume that percentageComplete is an integer, and not a floating point value)

src/components/ProgressBar/ProgressBar.less
```diff
@@ -0,0 +1,84 @@
...
+	&-bar-container {
+		overflow: hidden;
+		position: relative;
+		margin: @size-S 0;
+	}
+
+	&-bar {
+		background: @color-primary;
+		height: @size-XXS;
+		min-width: @bar-width;
```
chrisdopuch: Minimum width of 5000px seems really high?

PR #941, opened 2018-07-30
Fixes ANXR-1520 by adding diagonal tick labels to `Axis`

liliedelamotte:
## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [ ] Edge (Win 10)

### DIFF
--- a/src/components/Axis/Axis.jsx
+++ b/src/components/Axis/Axis.jsx
@@ -172,7 +172,9 @@ const Axis = createClass({
 					x =
 						textOrientation === 'vertical' ? 0 : orientationSign * tickSpacing;
 					y =
-						textOrientation === 'vertical' ? orientationSign * tickSpacing : 0;
+						textOrientation === 'vertical' || textOrientation === 'diagonal'
+							? orientationSign * tickSpacing
+							: 0;
 					dy =
 						textOrientation === 'vertical'
 							? '0em'
--- a/src/components/Axis/Axis.jsx
+++ b/src/components/Axis/Axis.jsx
@@ -74,6 +74,11 @@ const Axis = createClass({
 			an ordinal scale, this number is treated as an absolute number of ticks
 			to display and is powered by our own utility function \`discreteTicks\`.
 		`,
+
+		textOrientation: oneOf(['vertical', 'horizontal', 'diagonal'])`
+			Determines the orientation of the tick text. This may override what the orient prop
+			tries to determine.
+		`,
 	},
 
 	getDefaultProps() {
@@ -99,6 +104,7 @@ const Axis = createClass({
 			outerTickSize,
 			tickFormat = scale.tickFormat ? scale.tickFormat() : _.identity,
 			tickPadding,
+			textOrientation,
 			...passThroughs
 		} = this.props;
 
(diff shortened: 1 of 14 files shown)


src/components/Axis/Axis.jsx
```diff
@@ -108,6 +114,94 @@ const Axis = createClass({
 		const range = scale.range();
 		const sign = orient === 'top' || orient === 'left' ? -1 : 1;
 		const isH = orient === 'top' || orient === 'bottom'; // is horizontal
```
joelgriffith: Whooh this is a doozy

joelgriffith: Do you think it'd be _easier_ to supply a degree vs an enumarable value for textOrientation?

src/components/Axis/Axis.jsx
```diff
@@ -108,6 +114,94 @@ const Axis = createClass({
...
+				y,
+				dy,
+			};
+		};
+		const orientationProperties = {
+			vertical: getOrientationProperties(orient, 'vertical'),
+			horizontal: getOrientationProperties(orient, 'horizontal'),
+			diagonal: getOrientationProperties(orient, 'diagonal'),
+		};
+		const orientationKey = textOrientation || 'horizontal';
```
joelgriffith: We might want to list that this defaults to `horizontal` in the docs

src/components/Axis/Axis.jsx
```diff
@@ -147,16 +241,18 @@ const Axis = createClass({
...
 							className={cx('&-tick-text')}
-							x={isH ? 0 : sign * tickSpacing}
-							y={isH ? sign * tickSpacing : 0}
+							x={orientationProperties[orientationKey].x}
+							y={orientationProperties[orientationKey].y}
 							dy={
-								isH
-									? sign < 0 ? '0em' : '.71em' // magic d3 number
-									: '.32em' // magic d3 number
+								orientationProperties[orientationKey].dy
```
joelgriffith: Remove commented code?

src/components/Axis/Axis.jsx
liliedelamotte: Potentially. My only thinking is that the degrees we are using to rotate text here are not a clear 0, 45, and 90, instead it's 0, -30, and -90, which may make enumerable values the better option. Thoughts?

jondlm: I'm fine keeping it as enumerable values.

src/components/Axis/Axis.jsx
```diff
@@ -185,11 +186,7 @@ const Axis = createClass({
 				transform:
 					textOrientation === 'vertical'
 						? 'rotate(-90)'
-						: textOrientation === 'horizontal'
```
jondlm: Nice cleanup!

src/components/BarChart/BarChart.jsx
```diff
@@ -45,7 +46,7 @@ const BarChart = createClass({
 		MARGIN: {
 			top: 10,
 			right: 20,
-			bottom: 50,
+			bottom: 75,
```
jondlm: I don't think we want to change the default bottom margin since it should have already been tweaked to look good with horizontally oriented axis text.

src/components/BarChart/BarChart.jsx
```diff
@@ -387,10 +402,12 @@ const BarChart = createClass({
 				<g transform={`translate(${margin.left}, ${innerHeight + margin.top})`}>
 					<Axis
 						orient="bottom"
+						// textOrientation={xAxisTickCount > 2 ? "diagonal" : ''}
```
jondlm: Extra comment?

joelgriffith:
@liliedelamotte Looks like this branch is also out-of-date (plus Jon's requested changes). You'll likely need to merge master back into _this_ branch as well as update snapshots.

src/components/BarChart/BarChart.jsx
```diff
@@ -230,7 +241,7 @@ const BarChart = createClass({
 			margin: {
 				top: 10,
 				right: 20,
-				bottom: 50,
+				bottom: 75,
```
jondlm: Don't think we want to change this either.

joelgriffith:
@jondlm looks like the `bottom` issue is fixed, shall we commence merging? I can give it a run-through on them browsers as well

PR #947, opened 2018-08-13
Fixes #946 by making adjustments to `Button` padding

### LINKED ISSUE #946: Basic Button Icon + Text has uncentered padding
Buttons have standard padding, Icons have standard padding, it's looking like when we bundle all of that together, the icon + text buttons are lopsided, they have more space on the left side than on the right side. Is there something we can do to make the icon + text buttons even, without affecting the other buttons or other places where the checkIcon is used?
screenshot 2018-08-07 09 57 14

I feel like this came up before and we resolved it before...

liliedelamotte:
## PR Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] Edge (Win 10)

### DIFF
--- a/src/components/Button/Button.less
+++ b/src/components/Button/Button.less
@@ -40,10 +40,6 @@
 
 	&::before { border-radius: @Button-size-borderRadius; }
 
-	&&-has-only-icon {
-		padding: 0 5px;
-	}
-
 	& &-content {
 		display: flex;
 		justify-content: center;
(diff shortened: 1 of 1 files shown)


liliedelamotte:
Although the `Button` padding looks better than it did before, there's still a couple of problems. 
The first is that in the basic example, for buttons that have both text and an icon, there's a tiny bit more padding on the left side than the right (because of the `svg`). Is it noticeable enough to need change? I was having some difficulties removing that bit of space, as decreasing left padding affected more buttons than just those ones.
The second is that in the invisible example, the button with only an icon still retains padding on the right, whereas that padding has been removed on the buttons with only icons in the basic example. Not sure why this is happening to the invisible button. Suggestions on how to fix this?

jondlm:
Closing this one since I don't think it's quite the right solution to get all our buttons looking nice with icons.

PR #952, opened 2018-08-16
Fixes #946 by creating a new Button with Text and Icon

### LINKED ISSUE #946: Basic Button Icon + Text has uncentered padding

liliedelamotte:
## PR Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] Edge (Win 10)


src/components/ButtonWithIcon/ButtonWithIcon.jsx
```diff
@@ -0,0 +1,110 @@
...
+			isDisabled: false,
+			isActive: false,
+			onClick: _.noop,
+			type: 'button',
+		};
+	},
+
+	handleClick(event) {
+		const { isDisabled, onClick } = this.props;
+		const domNode = ReactDOM.findDOMNode(this);
```
jhsu: instead of finding a dom node everytime the buutton is clicked, you should store the `ref` of the thing you want to focus. See https://reactjs.org/docs/refs-and-the-dom.html

Also, right now, this focuses the `div`, is that intended? You probably want to store the ref of the `Button` so that the button is focused instead.

src/components/ButtonWithIcon/ButtonWithIcon.jsx
```diff
@@ -0,0 +1,110 @@
...
+			domNode.focus();
+			onClick({ event, props: this.props });
+		}
+	},
+
+	render() {
+		const {
+			isDisabled,
+			isActive,
+			iconIsOnLeft,
```
jhsu: i don't see `iconIsOnLeft` used

PR #957, opened 2018-10-11
Fixes #955 by allowing Bar component to accept both string and number type for dimension.

### LINKED ISSUE #955: Bar component doesn't support width and height in percentage unit
Bar component uses `<rect>` for its internal implementation. While `<rect>` accepts both length and percentage type for width and height, the propTypes of Bar only accepts length (number). This causes React to print console warnings if percentage (string) is passed to the component.

arun-xandr:
## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [ ] Edge (Win 10)
- ~One core team UX approval~

### DIFF
--- a/src/components/Bar/Bar.jsx
+++ b/src/components/Bar/Bar.jsx
@@ -41,11 +41,11 @@ const Bar = createClass({
 			y coordinate.
 		`,
 
-		height: number`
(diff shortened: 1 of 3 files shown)


arun-xandr:
Thanks @jondlm  👍
is there going to be any patch release planned in the near future ?

PR #963, opened 2018-12-14
Chore/update storybook etc

astropunk:
## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [ ] Safari
  - [ ] Edge (Win 10)


.storybook/lucid-docs-addon/PanelToggles.js
```diff
@@ -1,34 +1,39 @@
...
-	}
-
-	onToggleLayout() {
-		const urlState = this.props.api.getUrlState();
-		this.props.api.setOptions({
-			addonPanelInRight: !urlState.panelRight,
-		});
-	}
+const portalElement = window.document.createElement('div');
+window.parent.document.body.appendChild(portalElement);
```
astropunk: storybook puts content in an iframe, so the statically positioned toggle would be pinned to the iframe.  We shouldn't run into any cross-origin issues.

.storybook/lucid-docs-addon/PanelToggles.js
```diff
@@ -1,34 +1,39 @@
...
 
-	onTogglePanel() {
-		const urlState = this.props.api.getUrlState();
-		this.props.api.setOptions({
-			showAddonPanel: !urlState.addons,
-		});
-	}
+export const withPanelToggles = makeDecorator({
```
astropunk: Using storybook's decorator functionality since we only want the toggle visible on "example" pages and not our "article" pages.

scripts/path-import-modules.js
```diff
@@ -56,9 +56,7 @@ const exportCode = (specifierType, specifierPath, exportName) => {
 		`,
 	};
 
-	// For some reason this doesn't pickup our .babelrc even though it's supposed to
-	return babelCore.transform(codeMap[specifierType], { presets: ['es2015'] })
-		.code;
+	return babelCore.transform(codeMap[specifierType]).code;
```
astropunk: babel presets were removed.  Seems to pick up babelrc fine now.

src/components/Autocomplete/Autocomplete.spec.jsx
```diff
@@ -284,7 +284,7 @@ describe('Autocomplete', () => {
 			});
 
 			afterEach(() => {
-				if (wrapper) {
+				if (wrapper && wrapper.exists()) {
```
astropunk: fixes some flaky tests where we were trying to unmount components already unmounted.

src/components/Collapsible/__snapshots__/Collapsible.spec.jsx.snap
```diff
@@ -7,7 +7,9 @@ exports[`Collapsible [common] example testing should match snapshot(s) for 1.tog
       "height": 0,
     }
   }
-/>
+>
```
astropunk: enzyme shallow no longer uses a self closing tags, and now includes the immediate child's text/node

.storybook/lucid-docs-addon/PanelToggles.js
```diff
@@ -1,34 +1,39 @@
...
+			</div>
+		);
+	},
+});
 
+class PanelToggles extends React.Component {
 	render() {
-		const urlState = this.props.api.getUrlState();
-
-		if (urlState.full || !urlState.addons) {
```
astropunk: urlState.addons intended use is to control the visibility of the panel sidebar.  Using that here creates a bug where if the user hides the sidebar and later reloads the url we don't show the toggle component.  Example: https://docspot.adnxs.net/projects/lucid/latest/?selectedKind=DropMenu&selectedStory=stateless&full=0&addons=0&stories=1&panelRight=1&addonPanel=lucid-docs-panel-props

This code is effectively replaced by using the new panelToggle decorator.

jondlm:
Seems like a minor enough issue to me. I'd much rather have all these upgrades!

package.json
```diff
@@ -2,6 +2,9 @@
   "name": "lucid-ui",
   "version": "2.36.5",
   "description": "A UI component library from AppNexus.",
+  "engines": {
+    "node": ">=8.12.0"
```
astropunk: I think I need to remove this since it throws an error during npm install if the node engine doesn't match.  Is there a better way to flag that development must be done >=8, but it doesn't matter for consumers of the library?

jondlm: Good question. I've seen folks use a little bash script to check the version of node instead of `engines` but it's certainly not as clean. As far as I can tell there isn't a clear hook we could use to only run a script when running `yarn install` from within lucid (i.e. you're a developer working on lucid). Maybe we should just skip this problem for now and instead document it in the README?

PR #964, opened 2018-12-19
switch to babel.config.js so the path import script picks up babel settings

astropunk:
## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [ ] Safari
  - [ ] Edge (Win 10)


astropunk:
Some info on the switch from babelrc to babel.config.js.  Trying to think of any bad side-effects that might arise from this change...

https://new.babeljs.io/docs/en/next/configuration.html
https://babeljs.io/docs/en/config-files

Issue #974, opened 2019-04-05
Update close icon for Sidepanel


rsadres:
Use larger close icon in an invisible button in the Sidepanel header. Add more top and right padding to provide an easier target area to hit.

rsadres:
![Screenshot 2019-04-05 11 21 45](https://user-images.githubusercontent.com/10146234/55638395-107ecd80-5795-11e9-8f42-44f75a132e10.png)

miloofcroton:
I've made the containing element for the 'x' an invisible button.

We talked about making the 'x' greyed out, instead of teal, but it's already set to grey on our end. Adding an `!important` tag to overwrite whatever end users are doing. If this is heavy handed, we can pull it back.

Added top and right padding to the enclosing element too.

I want to make this better by having the path scale to fill that element, but I'm not sure how to do that without changing the path itself. This would minimize the increased minimum size of the header bar.

Feel free to check out, and I will try again after lunch: https://github.com/appnexus/lucid/tree/974-embiggen-close-button

Issue #968, opened 2019-01-30
Update close icon to have two sizes 


rsadres:
Please add a larger version of the close icon - two SVGs in a single component that you can switch between with a prop

rsadres:
[icon_close_big.svg.zip](https://github.com/appnexus/lucid/files/2814534/icon_close_big.svg.zip)

PR #976, opened 2019-04-10
#974, #968 - Larger Icon option and larger close button on the SidePanel header

miloofcroton:
## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [ ] Edge (Win 10)

### DIFF
--- a/src/components/Icon/CrossIcon/CrossIcon.jsx
+++ b/src/components/Icon/CrossIcon/CrossIcon.jsx
@@ -1,10 +1,13 @@
 import _ from 'lodash';
 import React from 'react';
+import PropTypes from 'react-peek/prop-types';
 import Icon from '../Icon';
 import { lucidClassNames } from '../../../util/style-helpers';
 import { createClass, omitProps } from '../../../util/component-types';
 
 const cx = lucidClassNames.bind('&-CrossIcon');
+
+const { bool } = PropTypes;
 
 const CrossIcon = createClass({
 	displayName: 'CrossIcon',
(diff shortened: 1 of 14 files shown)


src/components/Icon/CrossIcon/CrossIcon.jsx
```diff
@@ -22,18 +25,26 @@ const CrossIcon = createClass({
 
 	propTypes: {
 		...Icon.propTypes,
+		large: bool`
```
jondlm: I think it makes more sense to call this `size` and make it an enum like we do [over here](https://github.com/appnexus/lucid/blob/f5a89a2033935b52cf1f79966072461d9beeb333/src/components/Button/Button.jsx#L60). Just make it a `oneOf(['large'])` and make the default 16x16.

Side note, our convention is to us isX and hasX for boolean props. Those conventions are mentioned in [our contrib docs](https://github.com/appnexus/lucid/blame/f5a89a2033935b52cf1f79966072461d9beeb333/CONTRIBUTING.md#L63).

src/components/SidePanel/SidePanel.less
```diff
@@ -90,7 +91,9 @@
 						display: flex;
 						justify-content: center;
 						align-items: center;
-
+						padding: 8px;
+						margin: 12px;
```
jondlm: Make this `@size-standard` which is defined as `12px` in our [variables.less file](https://github.com/appnexus/lucid/blob/f5a89a2033935b52cf1f79966072461d9beeb333/src/styles/variables.less#L29).

src/components/Icon/CrossIcon/CrossIcon.jsx
miloofcroton: There's a problem with this because size is a prop that sometimes gets passed through to the generic Icon. What else could we call it?

jondlm: Oh dang, you're right. Good catch. How about `presetSize`?

miloofcroton: Sounds good to me

astropunk:
perfectly cromulent branch name.

Issue #977, opened 2019-04-11
Add section header to Tag


rsadres:
We have tag, and nested tag, can we please add a 3rd hierarchical level above nested. I'm not sure what to name it. Section header? Tag category?

- all caps section title
- blue text

rsadres:
![Screenshot 2019-04-11 11 22 30](https://user-images.githubusercontent.com/10146234/55969675-3819de00-5c4c-11e9-90ef-9465298b1db8.png)

jondlm:
Here's an example of what I'm thinking it would look like from the consumer perspective:

```
<Tag>
  GLOBAL:
  <Tag>
    Insertion Orders
    <Tag>Nikon - Universal - Mobile</Tag>
    <Tag>Nikon - US - Targeting</Tag>
  </Tag>
  <Tag>
    Traffickers
    <Tag>Bob</Tag>
  </Tag>
</Tag>

<Tag>
  IN-PROGRESS:
  <Tag>
    Transaction Recency
    <Tag>Nothing in the last 24 hours</Tag>
  </Tag>
</Tag>
```

I'm not 100% certain we'll be able to achieve it cleanly with the current component interface (without making any breaking changes) but it's worth a shot.

PR #978, opened 2019-04-12
#977 - 2nd order hierarchical tags are given a special style

miloofcroton:
We are now checking for 2nd order hierarchical tags and giving special style to them, with blue and all uppercase.

## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [ ] Edge (Win 10)


src/components/Tag/Tag.jsx
```diff
@@ -60,8 +60,15 @@ const Tag = createClass({
 		const { isRemovable, children, className, ...passThroughs } = this.props;
 
 		const subTags = filterTypes(children, Tag);
+
 		const otherChildren = rejectTypes(children, Tag);
+
 		const hasOtherChildren = !_.isEmpty(otherChildren);
+
+		const hasGrandChildren = !!subTags.find(
+			tag => parseInt(tag.key.slice(1, 2)) > 1
```
jondlm: I'm a little confused by this, what is `tag.key` and how is it determining that there are tag grand children? I'd imagine we'd need something like `filterTypes(subTags, Tag).length > 1` or something (not tested).

miloofcroton: The tags are from [React.children.toArray()](https://reactjs.org/docs/react-api.html#reactchildrentoarray) and the keys have a weird string representation, but if the second character is greater than 1, then it has a grandchild.

That's true, we could loop through the subTags too. That would be more readable too.

PR #983, opened 2019-05-10
full width sidepanel content

jhsu:
the sidepanel uses `display: flex` for the content, so make this content area `flex: 1` to be full width.


jondlm:
I think Nathan already did this on the `xandr` branch

jhsu:
yep he did.

PR #988, opened 2019-06-05
Additional expander label data and remove transition

KatsiarynaMashokha:
With a current implementation of **Expander**,  transition causes some flickering. 
Also in order to display any additional div or text on the same line as Expander Label, the div needs to be passed outside the **Expander** and then some absolute positioning needs to be applied to this div, as well as moving if to the right by 125px, and moving down the Expander Label to make sure, that they all are located on the same line.

Now you can pass an additional div, that you want to display on the same line with **AdditionalLabelContent** prop. Let me know pls if there is a better naming for it or some other comments regarding this PR.

Screen Shot 2019-06-04 at 6 34 22 PM
 

## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [ ] Edge (Win 10)

### DIFF
--- a/src/components/Expander/Expander.jsx
+++ b/src/components/Expander/Expander.jsx
@@ -1,7 +1,6 @@
 import _ from 'lodash';
 import React from 'react';
 import PropTypes from 'react-peek/prop-types';
-import ReactTransitionGroup from 'react-transition-group/CSSTransitionGroup';
 import { lucidClassNames } from '../../util/style-helpers';
 import {
 	createClass,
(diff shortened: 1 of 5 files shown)


jondlm:
Thanks for the PR! The change seems good to me from a high level. I'd like to loop in @rsadres to make sure she's on board as well from a UX perspective. Could you run `npm run upload-docs` so she could preview the branch on docspot?

KatsiarynaMashokha:
Ok, thank you. Just ran the command above:  Uploaded to: https://docspot.adnxs.net/projects/lucid/additional-expander-label-data

rsadres:
This looks good to me! 
these buttons near the expander are common to a lot of our pages, so it makes sense to me to build the ability into the component.
Chatted quickly with Kate, if you could add some more padding in the example - preferably to match that page template example, I'd appreciate it. This way we'll have an ideal we can point to for everyone to reference. 
thanks!

src/components/Expander/Expander.jsx
```diff
@@ -50,6 +49,22 @@ const Expander = createClass({
...
+			statics: {
+				peek: {
+					description: `
+						Renders a \`<span>\` to be shown next to the expander label.
+					`,
+				},
+			},
+			propName: 'AdditionalLabelContent',
+			propTypes: {
+				children: node`
```
astropunk: thoughts on this being a render prop function over a node?

jhsu: I don't think having this as a function would add any value, since we don't need `Expander` to inject anything into what we're rendering here.
We just want to provide content for `Expander` to place next to its label.

PR #993, opened 2019-07-02
first pass on custom createclass alternative

astropunk:
## PR Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] Edge (Win 10)

### DIFF
--- a/src/components/Expander/Expander.jsx
+++ b/src/components/Expander/Expander.jsx
@@ -2,179 +2,148 @@ import _ from 'lodash';
 import React from 'react';
 import PropTypes from 'react-peek/prop-types';
 import { lucidClassNames } from '../../util/style-helpers';
-import { createClass, findTypes, omitProps } from '../../util/component-types';
-import { buildHybridComponent } from '../../util/state-management';
+import { findTypes, omitProps, withReducer } from '../../util/component-types';
 import ChevronIcon from '../Icon/ChevronIcon/ChevronIcon';
 import Collapsible from '../Collapsible/Collapsible';
-import * as reducers from './Expander.reducers';
+import {initialState, reducer, mapDispatchToProps} from './Expander.reducers';
 
 const cx = lucidClassNames.bind('&-Expander');
 
 const { any, bool, func, node, object, oneOf, string } = PropTypes;
 
-const Expander = createClass({
-	displayName: 'Expander',
+const Expander = (props) => {
 
-	statics: {
-		peek: {
-			description: `
-				This is a container that provides a toggle that controls when the
-				content is shown.
-			`,
-			categories: ['layout'],
-			madeFrom: ['ChevronIcon'],
-		},
-	},
+	const {
+		children,
+		className,
+		style,
+		kind,
+		onToggle,
+		isExpanded,
+		...passThroughs
+	} = props;
 
-	components: {
-		Label: createClass({
-			displayName: 'Expander.Label',
-			statics: {
-				peek: {
-					description: `
-						Renders a \`<span>\` to be shown next to the expander icon.
-					`,
+	const labelChildProp = _.first(
+		_.map(findTypes(props, Expander.Label), 'props')
+	);
+
+	const additionalLabelContentChildProp = _.first(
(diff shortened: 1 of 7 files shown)


astropunk:
https://docspot.adnxs.net/projects/lucid/removecreateclass/?selectedKind=Expander&selectedStory=expanded&full=0&addons=1&stories=1&panelRight=1&addonPanel=lucid-docs-panel-props

astropunk:
- Existing Storybook and Peek documentation works.
- "Hybrid Component" state works in a similar manner.
    - No concept of `Selectors`, but that is currently only used in `Paginator`
    - Currently not handling the magic `InitialState` prop which allowed consumers to pass an InitialState prop that was used for default state on props controlled by Hybrid Components.  I don't believe this functionality was heavily used.  A more idiomatic approach moving forward might  be to let users compose their own HoC using `useReducer`, and pass an `initialState` meaningful to them.

src/components/Expander/examples/3.interactive-without-changing-labels.jsx
```diff
@@ -1,12 +1,13 @@
 import React from 'react';
 import createClass from 'create-react-class';
-import { Expander } from '../../../index';
+import { Expander, ExpanderStateful } from '../../../index';
 
+//@todo child components are kinda awkward since they won't be directly on HoC versions.  Should they be their own export?
```
astropunk: note^^

src/components/Expander/Expander.jsx
```diff
@@ -2,179 +2,148 @@ import _ from 'lodash';
...
+Expander.AdditionalLabelContent.propName = 'AdditionalLabelContent';
+Expander.AdditionalLabelContent.propTypes = {
+	children: node`
+		Used to display additional information or/and actions next to expander label.
+	`,
+};
+
+//@todo maybe just let consumers compose the HoC directly so they can pass their own initialState and whatnot?
+//@todo we could keep the stateful component as the default.  Just need to tweak our module loader in Storybook
+const ExpanderStateful = withReducer(Expander, reducer, mapDispatchToProps, initialState);
```
astropunk: ^^note

jondlm: The idea behind exporting the "hybrid component" as the default was because it allowed consumers to "take over" any of the props they needed without worrying too much. If that's a convention we break it's probably going to be a big change to get consumers to do the composition themselves. In a vacuum I'd agree that it's the right change to make though.

src/components/Expander/examples/3.interactive-without-changing-labels.jsx
jondlm: Again, in a vacuum I think the best approach would be to disconnect the child components from the root level component and just export them from the same file. Like `export { Expander, ExpanderLabel }` but it'd be a big lift for consumers to change their code unless we provided a codemod.

src/util/component-types.js
```diff
@@ -139,3 +139,15 @@ export function omitProps(props, type, keys = [], targetIsDOMElement = true) {
 			.concat(additionalOmittedKeys)
 	);
 }
+
+export function withReducer(WrappedComponent, reducer, mapDispatchToProps, initialState){
```
jondlm: Cool solution!

src/components/Expander/examples/3.interactive-without-changing-labels.jsx
astropunk: another option might be to have `withReducer` somehow tack on the child components on the wrapped HoC.

src/components/Expander/Expander.jsx
astropunk: yep the main difference here would be consumers either opt-in or opt-out of driving props controlled by the reducers.  With Hybrid Components consumers could use the stateful version but still drive the props themselves and the internal state would be ignored.  I'm not convinced the existing behavior is ideal since it leads to an amount of magic if you don't fully grok what is happening.  
 
The lift for consumers as it stands here would be searching for the 22 Hybrid Components where they also pass in a prop controlled by Hybrid state.

jhsu:
could you maybe update the description of this to outline the objectives?

PR #994, opened 2019-07-03
Demo of the react-js-to-ts tool

miloofcroton:
1. I replaced react-peep/prop-types with the official prop-types package.
2. I added `PropTypes.` in front of every prop type before running the conversion, as opposed to getting them off the destructured PropTypes object at the top. Otherwise, the tool doesn't recognize these functions.
3. I removed the comments from prop type definitions because they make the tool fail
4. PropTypes do not convert properly for child components. Pulling the components out doesn't work either. 

The resulting file is the what we have.

Potential improvements:

- Make the tool fix prop types automatically (steps 1 & 2) for us?
- Make the prop type definitions automatically convert into comments that we can use with Storybook in the future?
- How to fix child components? Should we export { Expander, Label, AdditionalContentLabel } instead? If so, we can completely move them outside of the original function. If not, we'll either have to manually set types for them in TypeScript, or perhaps we're lucky enough that we can add that functionality to the tool too.


src/components/Expander/Expander.jsx
```diff
@@ -23,6 +21,20 @@ const Expander = (props) => {
 		...passThroughs
 	} = props;
 
+	Expander.Label = () => null;
```
astropunk: we wouldn't want to redefine all this stuff every render tho?

src/components/Expander/Expander.jsx
```diff
@@ -1,6 +1,6 @@
 import _ from 'lodash';
 import React from 'react';
-import PropTypes from 'react-peek/prop-types';
+import PropTypes from 'prop-types';
```
astropunk: did it not work with peek?   We'll have to figure out what sequence of refactor tasks allow docs to work during every step.

src/components/Expander/Expander.jsx
miloofcroton: True. Was just trying to get the tool to work. It fails if it's in a separate code block because the functions below cause errors.

```
  const labelChildProp = _.first(
     _.map(findTypes(props, Expander.Label), "props")
   );
   const additionalLabelContentChildProp = _.first(
     _.map(findTypes(props, Expander.AdditionalLabelContent), "props")
   );
```

I also can't put them above Expander in separate code blocks because then I'm attacking Label to Expander which doesn't exist yet.

src/components/Expander/Expander.jsx
miloofcroton: Nope, didn't work with peek

PR #997, opened 2019-07-10
fix SidePanel's content heights 

jhsu:
This fixes SidePanel scrolling is able to show all content, even with top offset.

previously, the sidepanel would bleed below the viewport resulting in content that went offscreen.

The fix involves using `height: auto` and setting top, bottom and left/right to 0, since the panel is `display: fixed;`. This allows a margin-top to be applied without pushing it below the viewport.

working version:
working version

## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [ ] Edge (Win 10)

### DIFF
--- a/src/components/SidePanel/SidePanel.less
+++ b/src/components/SidePanel/SidePanel.less
@@ -6,10 +6,11 @@
 
 	.@{prefix}-SidePanel-pane {
 		position: fixed;
+		height: auto;
 		top: 0;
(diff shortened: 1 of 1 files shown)


jhsu:
not sure why coverage says it dropped

njlg:
Maybe the percent never updated after the last big merge ?

PR #999, opened 2019-07-11
default textfield autocomplete to off.  Requested because with wrappe…

astropunk:
…rs like SingleSelect it can be tedious to pass the autocomplete prop through

## PR Checklist

  - [x] Chrome
  - [x] Firefox

### DIFF
--- a/src/components/TextField/TextField.jsx
+++ b/src/components/TextField/TextField.jsx
@@ -90,6 +90,10 @@ const TextField = createClass({
 			heavily inspired by the
 			[lazy-input](https:/docs.npmjs.com/package/lazy-input) component.
 		`,
+
+		autoComplete: string`
+			html attribute for controlling browser autocomplete functionality
+		`
 	},
 
 	getDefaultProps() {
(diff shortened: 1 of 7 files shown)


astropunk:
we ok with this as a semver patch?

src/components/TextField/TextField.jsx
```diff
@@ -105,6 +109,7 @@ const TextField = createClass({
 			debounceLevel: 500,
 			lazyLevel: 1000,
 			value: '',
+			autoComplete: 'off',
```
jondlm: Seems to me like we should probably enable it by default and only default disable it for cases like drop menu and other selects. I just wonder what impact this might have on users who do actually care about the default browser's autocomplete for text fields.

astropunk: Ok sure makes sense to me.  `SearchableMultiSelect`, `SearchableSelect`, `SearchableSingleSelect` -- any others I'm missing?   Can you think of any situations where these components would want autocomplete?  I'm trying to decide if we want it to be drivable by a prop.

PR #1004, opened 2019-07-16
Typography component

miloofcroton:
First pass is done. I can think of a ton of features we could add, but it could expand the API needlessly. I would say we should poll the library consumers to see the most common styles that are added to generic text components, so maybe we could generate those via props instead (allowing for DRYer code).

Docspot: https://docspot.adnxs.net/projects/lucid/typography-component/?selectedKind=Typography&selectedStory=variants&full=0&addons=1&stories=1&panelRight=1&addonPanel=lucid-docs-panel-props

## PR Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] Edge (Win 10)

### DIFF
--- a/src/components/Typography/Typography.jsx
+++ b/src/components/Typography/Typography.jsx
@@ -13,7 +13,7 @@ const Typography = createClass({
 	statics: {
 		peek: {
 			description: `
-				A general component for several types of semantic textual HTML elements.
+				A general component for several types of textual HTML elements.
 			`,
 			categories: ['text'],
 		},
@@ -29,7 +29,8 @@ const Typography = createClass({
 		`,
 
 		variant: oneOf(['h1', 'h2', 'h3', 'p', 'a', 'pre', 'code'])`
-			This prop defines the type of HTML element that will be displayed.
+			This prop defines the type of text that will be displayed.
+			It may be an actual HTML element or something with extra semantic meaning.
 		`,
 	},
 
@@ -39,18 +40,39 @@ const Typography = createClass({
 		};
 	},
 
+	elFinder: variant => {
+		switch (variant) {
+			case 'h1':
+				return 'h1';
+			case 'h2':
+				return 'h2';
+			case 'h3':
+				return 'h3';
+			case 'p':
+				return 'p';
+			case 'a':
+				return 'a';
+			case 'pre':
+				return 'pre';
+			case 'code':
+				return 'code';
+			case 'tabular':
+				return 'p';
+			default:
+				return 'p';
+		}
+	},
+
 	render() {
-		const {
-			children,
-			className,
-			variant: Element,
-			...passThroughs
-		} = this.props;
+		const { children, className, variant, ...passThroughs } = this.props;
+		const { elFinder } = this;
+
+		const Element = elFinder(variant);
 
 		return (
 			<Element
 				{...omitProps(passThroughs, Typography)}
-				className={cx('&', `&-${Element}`, className)}
+				className={cx('&', `&-${variant}`, className)}
 			>
 				{children}
 			</Element>
--- a/src/components/Typography/Typography.jsx
+++ b/src/components/Typography/Typography.jsx
@@ -40,10 +40,12 @@ const Typography = createClass({
 		h2: 'h2',
 		h3: 'h3',
 		p: 'p',
+		tabular: 'p',
 		a: 'a',
 		pre: 'pre',
 		code: 'code',
-		tabular: 'p',
+		kbd: 'kbd',
+		samp: 'samp',
 	},
 
 	render() {
--- a/src/components/Typography/Typography.jsx
+++ b/src/components/Typography/Typography.jsx
@@ -28,19 +28,29 @@ const Typography = createClass({
 			Appended to the component-specific class names set on the root element.
 		`,
 
-		variant: oneOf(['h1', 'h2', 'h3', 'p', 'a', 'pre', 'code', 'tabular'])
-			.isRequired`
+		variant: oneOf([
+			'p',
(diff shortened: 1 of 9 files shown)


src/components/Typography/Typography.jsx
```diff
@@ -0,0 +1,83 @@
...
+	},
+
+	getDefaultProps() {
+		return {
+			variant: 'p',
+		};
+	},
+
+	elFinder: variant => {
+		switch (variant) {
```
jhsu: I'd suggest using an object instead of a switch for faster lookup.

jhsu:
bitmoji

astropunk:
for future reviews would you mind running the `upload-docs` npm script and posting a link in the review?  Saves a bit of time with loading up the project to take a look.  We should add it the PR template.  I forget to do it sometimes too.

src/components/Typography/Typography.jsx
```diff
@@ -0,0 +1,83 @@
...
+	propTypes: {
+		children: node`
+			Can contain elements or nested \`Typography\` components.
+		`,
+
+		className: string`
+			Appended to the component-specific class names set on the root element.
+		`,
+
+		variant: oneOf(['h1', 'h2', 'h3', 'p', 'a', 'pre', 'code'])`
```
astropunk: I think I'd like this to be required over having a default of `p`.  Thoughts?

src/components/Typography/Typography.less
```diff
@@ -0,0 +1,54 @@
...
+		font-weight: @font-weight-medium;
+		color: @color-linkColor;
+		text-decoration: none;
+	}
+	&-a:hover {
+		color: @color-linkColorHover;
+		text-decoration: underline;
+	}
+
+	// NOTE: kbd and samp are currently not prop options for `variant`
```
astropunk: let's just remove these or add them as variants then?

src/components/Typography/Typography.jsx
```diff
@@ -0,0 +1,83 @@
...
+				return 'h3';
+			case 'p':
+				return 'p';
+			case 'a':
+				return 'a';
+			case 'pre':
+				return 'pre';
+			case 'code':
+				return 'code';
+			case 'tabular':
```
astropunk: doesn't look like `tabular` is a variant prop.  If it's not a variant it doesn't seem like we need this fn?

src/components/Typography/Typography.jsx
jondlm: I second that.

src/components/Typography/Typography.jsx
jondlm: Also, I think [`<data>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/data) might be better for tabular data

src/components/Typography/Typography.jsx
miloofcroton: Good idea

src/components/Typography/Typography.jsx
miloofcroton: I like that better too

src/components/Typography/Typography.less
miloofcroton: Seems super simple to just add them. I left them out because the JIRA ticket didn't mention them.

src/components/Typography/Typography.jsx
miloofcroton: <data> could be the native element we use. I didn't want to use `data` as the key, just in case that we wanted to do something else with it.

miloofcroton:
> for future reviews would you mind running the `upload-docs` npm script and posting a link in the review? Saves a bit of time with loading up the project to take a look. We should add it the PR template. I forget to do it sometimes too.

For sure

rsadres:
What is preformatted text? and how is it different than code?
Everything else looks good to me.

miloofcroton:
I feel like this is a big opportunity to add some features that are used frequently. What might our consumers actually use regularly? If everything is too custom to add in, then fine. But if they're just adding basic things like color or bold or underline or kerning, maybe it would be handy to have those as props.

rsadres:
Good point. Also a good opportunity to limit. We only license 2 font weights so we should only offer those. 

I think users commonly create field labels, field values, and table data. All the other parts are incorporated into composite components like Header inside Dialog.

src/components/Typography/Typography.less
```diff
@@ -0,0 +1,54 @@
...
+		font-size: 16px;
+		line-height: 20px;
+	}
+	&-h3 {
+		font-size: 12px;
+		line-height: 16px;
+		text-transform: uppercase;
+		color: @color-neutral-6;
+	}
+
```
jhsu: there doesn't seem to be a style for `&-p`, should there be?

src/components/Typography/Typography.less
```diff
@@ -0,0 +1,58 @@
...
+	&-h1,
+	&-h2,
+	&-h3 {
+		margin: 0;
+		font-weight: @font-weight-medium;
+		text-rendering: optimizelegibility;
+	}
+	&-h1 {
+		font-size: 24px;
+		line-height: 28px;
```
dsusskind: Do we want all these explicit line-heights or can we rely on master.less to set it to 1.25? These px values seem pretty close to 1.25 x font-size

miloofcroton: I just pulled in the master.less values for h1

src/components/Typography/Typography.less
miloofcroton: I added an explicit statement in the less file for it. Not much to it, but it's there.

PR #1013, opened 2019-07-29
Typescript build

miloofcroton:
## PR Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] Edge (Win 10)


src/components/Collapsible/Collapsible.tsx
```diff
@@ -146,7 +175,7 @@ const Collapsible = createClass({
 						rootType,
 						{
 							...omitProps(passThroughs, Collapsible),
-							ref: this.storeRef('root'),
+							ref: (ref: any) => (this.rootRef = ref),
```
dsusskind: can we avoid using `any` here? looks like in other components like Button we didn't need to specify the type

miloofcroton: This looks like solid info on what `ref`s actually are: https://medium.com/@martin_hotell/react-refs-with-typescript-a32d56c4d315

tests/init.js
```diff
@@ -1,11 +1,12 @@
-import _ from 'lodash';
+// // Mock RAF
```
jondlm: Remove

Issue #1009, opened 2019-07-25
Add drag-able list component


rsadres:
There are a few places in our product where a drag-able list is needed in order to allow users to define the order.

Please add the one-off drag-able list you created to the library for re-use

rsadres:
merged

PR #1016, opened 2019-08-01
Add component to render a drag and drop list

krishoyt:
## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [x] Edge (Win 10)

Adds a basic list component that allows rearranging items by drag and drop: https://github.com/appnexus/lucid/issues/1009.

Jul-31-2019 15-15-56

### DIFF
--- a/src/components/DraggableList/DraggableList.jsx
+++ b/src/components/DraggableList/DraggableList.jsx
@@ -131,7 +131,7 @@ const DraggableList = createClass({
 
 		if (this.lastItemEl) {
 			const { bottom } = this.lastItemEl.getBoundingClientRect();
-			if (_.isNumber(dragIndex) && event.clientY > bottom) {
+			if (_.isFinite(dragIndex) && event.clientY > bottom) {
 				onDragOver(childCount, { event, props: this.props });
 			}
 		}
--- a/src/components/DraggableList/DraggableList.jsx
+++ b/src/components/DraggableList/DraggableList.jsx
@@ -1,0 +1,226 @@
+import _ from 'lodash';
+import React from 'react';
+import PropTypes from 'react-peek/prop-types';
+import { lucidClassNames } from '../../util/style-helpers';
+import { createClass, findTypes, omitProps } from '../../util/component-types';
+import { buildHybridComponent } from '../../util/state-management';
+import DotsIcon from '../Icon/DotsIcon/DotsIcon';
+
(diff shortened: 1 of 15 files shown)


rsadres:
Looking good!
Can you please make a few styling changes?
- make the draggable list item border #e8e6e6
- make the draggable list item border on hover it can stay what you have it right now #d3d1d1
- Please change the draggable list item handle to be #587eba at 50%
- Please change the draggable list item handle on hover to be #587eba at 100%
- Please use a gripper icon for draggable list item handle - rather than two dots icons which is what you have right now. 
- Please update padding-left to 9px for draggable list item so that text is indented a bit more

Thank you!

src/components/DraggableList/DraggableList.jsx
```diff
@@ -0,0 +1,226 @@
...
+		};
+	},
+
+	handleDragLeave(event) {
+		const { dragIndex, onDragOver } = this.props;
+		const childCount = findTypes(this.props, DraggableList.Item).length;
+
+		if (this.lastItemEl) {
+			const { bottom } = this.lastItemEl.getBoundingClientRect();
+			if (_.isNumber(dragIndex) && event.clientY > bottom) {
```
astropunk: if we're ever worried about guarding against NaN in these isNumber checks we could switch to isFinite

krishoyt: Sure, I'll change it.

src/components/DraggableList/DraggableList.jsx
```diff
@@ -0,0 +1,226 @@
...
+						'&-is-dragging': _.isNumber(dragIndex),
+					},
+					className
+				)}
+				style={style}
+				onDragLeave={this.handleDragLeave}
+			>
+				{_.map(itemChildProps, (itemChildProp, index) => {
+					return (
+						<div key={index}>
```
jondlm: I haven't thought too hard about it, but could using the index as the `key` cause some mis-renders in this component if the order of items changes in the source data?

krishoyt: That's a really good question.  I hate using indexes for keys at all, but it gets a little tricky when you have unknown children.  I see using indexes as keys a lot in this code base, but this is kind of a special case because you will actively be reordering the children.  

Do we have any other patterns for handling this type of situation?  I could maybe append whatever `getCombinedChildText` returns for each item?  Not sure if that would cause performance issues though.

PR #1035, opened 2019-09-04
pass the data point to the xFormatter and yFormatter functions

jhsu:
when formatting x or y for a bar, it can be useful for the formatter function to have access to the full data object.

This is useful when the bar is a sum of a range of data that has been grouped. for example a datum might look like: `{x: 1, y: 10, min: 0, max: 2}` we may want to format the display to show `0-2`.


jondlm:
Nice change! Looks good to me.

PR #1048, opened 2019-09-19
adjust table header's border and style behavior

astropunk:
## PR Checklist

  - [x] Chrome
  - [x] Firefox

### DIFF
--- a/src/components/Table/Table.less
+++ b/src/components/Table/Table.less
@@ -115,13 +115,16 @@
 		background-color: @color-table-header-dark-background;
 		color: @color-table-header-light-color;
 
-		// Table header cell styles
-		& > .@{prefix}-Table-Tr > .@{prefix}-Table-Th {
-			border-bottom: 1px solid @color-table-header-border;
-			&.@{prefix}-Table-has-light-header {
+		&:hover {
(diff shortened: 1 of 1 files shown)


src/components/Table/Table.less
```diff
@@ -153,6 +156,10 @@
 				opacity: 1;
 			}
 
+			&.@{prefix}-Table-is-resizing > .@{prefix}-Table-Th-inner > .@{prefix}-Table-Th-inner-resize {
```
astropunk: stops the resizer from losing hover and growing/shrinking as it is dragged around

src/components/Table/Table.less
```diff
@@ -182,7 +189,7 @@
 				.@{prefix}-Table-Th-inner-resize {
 					cursor: ew-resize;
 					width: 3px;
```
astropunk: I tried adding a negative right margin so that the resizer was centered over the border.  Worked in FF, but the hover states were weird in Chrome.

astropunk:
https://docspot.adnxs.net/projects/lucid/table-header-style-changes/?path=/story/components-table-datatable--resizable-fixed-headers-table

PR #1052, opened 2019-09-27
set padding-right on SearchField

jhsu:
## Problem:

search texts overlaps the search icon on the `SearchField` component

Screen Shot 2019-09-27 at 12 58 03 PM

## Solution:

add a padding-right to the input

Screen Shot 2019-09-27 at 1 13 42 PM

### DIFF
--- a/src/components/TextField/TextField.less
+++ b/src/components/TextField/TextField.less
@@ -18,8 +18,6 @@
 
 	&:focus {
 		border: @size-borderWidthFocused solid @color-primary;
-		padding-left: @size-S;
-		padding-right: @size-S;
 		outline: none;
 		border-color: @color-primary;
 	}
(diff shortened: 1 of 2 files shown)


src/components/TextField/TextField.less
```diff
@@ -18,8 +18,6 @@
 
 	&:focus {
 		border: @size-borderWidthFocused solid @color-primary;
-		padding-left: @size-S;
-		padding-right: @size-S;
```
jhsu: the default TextField already uses this for padding, so i removed it here so that we could overwrite it in SearchField

jondlm: Does it cause the text to jump around in the input when focus is applied?

jhsu: nope it doesn't

jondlm: Oh yeah, you're right. Cool looks good.

PR #1060, opened 2019-10-24
adjust hover colors for SuccessLightIcon and MinusCircleLightIcon

astropunk:
Xandr employees can preview changes [here](https://docspot.adnxs.net/projects/lucid/light-icon-hover/?path=/docs/icons-icons-successlighticon--default)

  - [x] Chrome
  - [x] Firefox
  - [ ] Safari
  - [ ] Edge (Win 10)

### DIFF
--- a/src/components/Icon/MinusCircleIcon/MinusCircleIcon.less
+++ b/src/components/Icon/MinusCircleIcon/MinusCircleIcon.less
@@ -14,7 +14,7 @@
 	}
 
 	&-is-clickable:not(&-is-disabled):hover &-background {
-		stroke: shade(@color-secondary-1, 20%);
-		fill: shade(@color-secondary-1, 20%);
+		stroke: @color-secondary-1-hover;
+		fill: @color-secondary-1-hover;
 	}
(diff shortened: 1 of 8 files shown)


dsusskind:
Should it be filled in when isActive? I'm seeing isActive basically behave the same as isClickable in the docspot link

astropunk:
> Should it be filled in when isActive? I'm seeing isActive basically behave the same as isClickable in the docspot link

Hmm, my understanding is that `isActive` should be the same effect as `isClickable` except without the cursor change on hover. @rsadres do you have any input?

rsadres:
I think I agree with Dan. I think isActive should be filled in. The only difference should be the hover color of isClickable as opposed to the hover color of isActive isClickable.
I think.
R

astropunk:
I added a new icon to the example that is clickable and toggles isActive.  Does that feel right to you @rsadres ?

https://docspot.adnxs.net/projects/lucid/light-icon-hover/?path=/docs/icons-icons-successlighticon--default

PR #1066, opened 2019-11-08
[proposal] annotate Dialog FC prop argument

jhsu:
When using lucid components in a typescript application, there are errors when using functional components that have default props.

annotating react function components are weird with typescript and defaultProps.

Instead of casting props as a type inside the function, annotate the prop argument and remove `React.FC` to get the correct prop annotation that takes `defaultProps` into account when rendering the component.

see example of `React.FC` vs annotating the prop argument.:
https://codesandbox.io/s/relaxed-cori-6cyds

typescript issue: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/30695

### DIFF
--- a/src/components/Dialog/Dialog.tsx
+++ b/src/components/Dialog/Dialog.tsx
@@ -90,7 +90,9 @@ export interface IDialogFC extends FC<IDialogProps> {
 	Footer: FC<IDialogFooterProps>;
 }
 
-export const Dialog: IDialogFC = (props): React.ReactElement => {
+export const Dialog = (
+	props: FixDefaults<IDialogProps, typeof defaultProps>
+) => {
 	const {
 		className,
 		size,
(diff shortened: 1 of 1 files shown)


jondlm:
Just wanted to drop a message here saying I'm still looking into this one. I haven't yet been able to figure out a way to make your proposal work without losing all the type safety on the actual functional component itself. I'm starting to feel like the quote is right:

> Unfortunately however it seems there's no way forward for both FunctionComponent and a well typed defaultProps.

https://github.com/DefinitelyTyped/DefinitelyTyped/issues/30695#issuecomment-474780159

I'll keep digging to see what can be done or which trade-off is less bad for us.

jondlm:
[This codesandbox] is the best I've been able to come up with so far that doesn't lose the type safety of the properties we attach to the function. It's not perfect because it requires us to be diligent about using the proposed `MakeDefaultsOptional` utility type.

jondlm:
Okay, here's another idea I had that I think I like the best: #1074

jhsu:
yeah, I think annotating the FC prop argument is the best route. Ensuring that consumers don't get errors should be the top priority.

jhsu:
closing as this has been discussed with @jondlm

PR #1070, opened 2019-11-12
Selection component typescript conversion

dreidn:
## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [ ] Edge (Win 10)

Docspot Lucid Docs:
https://docspot.adnxs.net/projects/lucid/selection-typescript-conversion/?path=/docs/communication-selection--kinds


src/components/Selection/Selection.tsx
```diff
@@ -0,0 +1,305 @@
...
+const defaultProps = {
+    isRemovable: true,
+    onRemove: _.noop,
+    hasBackground: false,
+    isBold: false,
+    kind: 'default' as SelectionKind,
+    responsiveMode: 'large' as SelectionResponsiveMode
+}
+
+const Selection: ISelectionFC = (props): React.ReactElement => {
```
jhsu: you shouldn't need to annotate the return type when you have `ISelectionFC`

dreidn: that makes sense, good catch

src/components/Selection/Selection.tsx
```diff
@@ -0,0 +1,305 @@
...
+Selection.displayName = 'Selection';
+Selection.Icon = SelectionIcon;
+Selection.Label = SelectionLabel;
+Selection.peek = {
+            description: `
+                Used to indicate selections. It's very similar to \`Tag\` but is meant
+                to be used in areas of the UI that have more space available to them.
+            `,
+            categories: ['communication'],
+        }
```
yasskin: More of a UX issue: The way this text shows up in Storybook "Notes" could be clearer. Because Selection is off the the left so far.  "It's" lacks a clear referent, so a fix could be something like: `Used to indicate selections. Selection is very similar to \`Tag\` but` ...
Screen Shot 2019-11-20 at 5 45 00 PM

docs/index.stories.js
```diff
@@ -46,9 +46,9 @@ const getDefaultExport = module => {
 const isPrivate = component =>
 	component._isPrivate || (component.peek && component.peek.isPrivate);
 
-const getExamplesFromContext = (reqExamples, rawContext) =>
+const getExamplesFromContext = (reqExamples, rawContext) => 
 	_.map(loadAllKeys(reqExamples, rawContext), ({ key, module, raw }) => ({
-		name: _.join(_.reject(_.words(key), w => /^(\d+)|jsx?$/.test(w)), ' '),
+		name: _.join(_.reject(_.words(key), w => /^(\d+)|^(j|t)sx?$/.test(w)), ' '),
```
jondlm: I think I already fixed this on master. It was tricky to get right 😧

src/components/Selection/Selection.tsx
```diff
@@ -0,0 +1,299 @@
...
+     * */
+    responsiveMode: SelectionResponsiveMode,
+} 
+
+const defaultProps = {
+    isRemovable: true,
+    onRemove: _.noop,
+    hasBackground: false,
+    isBold: false,
+    kind: 'default' as SelectionKind,
```
jondlm: I'd actually use `as const` instead since they tells TS that it's a hard coded value that won't change and TS will make sure it fits correctly otherwise you could end up with something bad like this: https://www.typescriptlang.org/play/index.html#code/C4TwDgpgBAYg9nKBeKByOA7CqoB83ADucqA3AFDkCWGwEATgGYCGAxtAAr1xgDOUAb3JQRURggBcsBBQC+lVpl7AodZckHDR4uFNQYEkbFGb94ccrIrlFGdWuAAmKVx78UD0kA

jondlm: These are actually better suited using `as const` which tells TS that the value is hard coded and can't be changed. Otherwise you can end up with something bad like this that type checks but is wrong: https://www.typescriptlang.org/play/index.html#code/C4TwDgpgBAYg9nKBeKByOA7CqoB83ADucqA3AFDkCWGwEATgGYCGAxtAAr1xgDOUAb3JQRURggBcsBBQC+lVpl7AodZckHDR4uFNQYEkbFGb94ccrIrlFGdWuAAmKVx78UD0kA

src/util/component-types.ts
```diff
@@ -15,6 +15,8 @@ export interface StandardProps {
 	children?: React.ReactNode;
 	/** Styles that are passed through to native control. */
 	style?: React.CSSProperties;
+
+    callbackId?: string,
```
jondlm: Formatting looks a little weird here. Are you using prettier to format stuff on save? I'm sure there's lots of vscode stuff for that.

docs/index.stories.js
dreidn: awesome, I'll use whats on master

src/components/Selection/Selection.tsx
dreidn: Oh that's really good to know. I'll `as const` const here

src/util/component-types.ts
dreidn: That does look strange. I was lazy about setting up my IDE and was just running the command, and I probably missed this file. Definitely easier to just do it on save so I'll set that up.

src/components/Selection/Selection.tsx
dreidn: Good point, I agree that is more clear. I'll change this to say `Selection is...`

PR #1073, opened 2019-11-18
Convert vertical tabs

yasskin:
## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [x] Edge (Win 10)

Here are the docspot links:

Storybook:
https://docspot.adnxs.net/projects/lucid/convert-vertical-tabs/

Vertical Tabs:
https://docspot.adnxs.net/projects/lucid/convert-vertical-tabs/?path=/docs/navigation-verticaltabs--title-as-prop


src/components/VerticalTabs/VerticalTabs.tsx
```diff
@@ -0,0 +1,202 @@
...
+				: selectedIndex;
+
+		return (
+			<div
+				{...omitProps(passThroughs, undefined, _.keys(VerticalTabs.propTypes))}
+				className={cx('&', className)}
+			>
+				<VerticalListMenu
+					selectedIndices={[
+						_.isUndefined(actualSelectedIndex) ? 0 : actualSelectedIndex,
```
yasskin: An example I saw in the code base for a similar index had "" instead of a zero; the default is 0, and type is number, so a zero seemed more precise, even though "" is == a truthy zero (but not === 0)

src/components/VerticalTabs/VerticalTabs.tsx
```diff
@@ -0,0 +1,202 @@
...
+									'props.children',
+									''
+								)}
+							</span>
+						</VerticalListMenu.Item>
+					))}
+				</VerticalListMenu>
+				<div className={cx('&-content')}>
+					{_.get(tabChildProps, [
+						_.isUndefined(actualSelectedIndex) ? 0 : actualSelectedIndex,
```
yasskin: I did the same thing here. Strictly equals zero as instead of `_.isUndefined(actualSelectedIndex) ? "" : actualSelectedIndex,`

src/components/VerticalTabs/VerticalTabs.tsx
```diff
@@ -0,0 +1,206 @@
...
+Tab.propName = 'Tab';
+Tab.propTypes = {
+	isSelected: bool`
+        Determines if the Tab is selected.
+    `,
+};
+
+/** Vertical Tabs Title Child Component */
+interface IVerticalTabsTitleProps extends StandardProps {}
+
```
yasskin: Should we declare empty interfaces? 

We could just declare:
`const Title = (_props: StandardProps): null => null; `

This line produces an es-lint warning: 46:11  warning  An interface declaring no members is equivalent to its supertype  @typescript-eslint/no-empty-interface.

This rule aims to ensure that only meaningful interfaces are declared in the code. The error is discussed here:

https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-empty-interface.md

We're not getting any type safety from the IVerticalTabsTitleProps interface, and, in any case, StandardProps only has optional types.
```
export interface StandardProps {
	/** Appended to the component-specific class names set on the root element.
			Value is run through the `classnames` library. */
	className?: string;
	/** Any valid React children. */
	children?: React.ReactNode;
	/** Styles that are passed through to native control. */
	style?: React.CSSProperties;
}
```

jondlm: I'm fine moving away from these empty interfaces. We never had a strong reason for keeping them other than simply getting a different name for the same thing.

src/components/VerticalTabs/VerticalTabs.reducers.spec.js
```diff
@@ -1,5 +1,5 @@
 import assert from 'assert';
-import * as reducers from './VerticalTabs.reducers.js';
```
jondlm: Good catch. We shouldn't have extensions on our imports.

src/components/VerticalTabs/VerticalTabs.tsx
```diff
@@ -0,0 +1,209 @@
...
+interface IVerticalTabsPropsRaw extends StandardProps {
+	/** Custom Tab component (alias for `VerticalTabs.Tab`) */
+	Tab?: React.ReactNode;
+
+	/** Custom Title component (alias for `VerticalTabs.Title`) */
+	Title?: React.ReactNode;
+
+	/** Indicates which of the \`VerticalTabs.Tab\` children is currently
+    selected */
+	selectedIndex?: number;
```
jondlm: I think this should be marked as required since it has a default prop

src/components/VerticalTabs/VerticalTabs.tsx
```diff
@@ -0,0 +1,209 @@
...
+		`,
+	};
+
+	static defaultProps = defaultProps;
+	static reducers = reducers;
+	static Tab = Tab;
+	static Title = Title;
+
+	// For backward compatibility with buildHybridComponent
+	static definition = {
```
jondlm: You can delete this `definition` static and simply move the `peek` property to a new, top-level static instead as seen here: https://github.com/appnexus/lucid/blob/de3aa74d866a04c4d6a09f591f5c753f855dae76/src/components/ToolTip/ToolTip.tsx#L145

src/components/VerticalTabs/VerticalTabs.tsx
jondlm: Isn't this check redundant with the default prop though?

src/components/VerticalTabs/VerticalTabs.tsx
yasskin: I removed the empty interface

src/components/VerticalTabs/VerticalTabs.tsx
yasskin: Done! Thx!

src/components/VerticalTabs/VerticalTabs.tsx
yasskin: done!

src/components/VerticalTabs/VerticalTabs.tsx
yasskin: Good point. It is redundant. I removed it.

src/components/VerticalTabs/VerticalTabs.tsx
yasskin: Great. I took it out.

src/components/VerticalTabs/VerticalTabs.tsx
yasskin: Done! Good catch.

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
jondlm: Interesting. What is this file?

jondlm:
Feel free to ignore the codecov check since it's a tiny dip

.vscode/settings.json
dreidn: @yasskin If you needed to add this setting, you may be able to add it to your IDE's settings instead. I think adding settings here will enable them for everyone running the project in VSCode.

PR #1074, opened 2019-11-19
Potential fix for FC type signatures

jondlm:
This moves the type checking for additional props we tack onto
functional components to our runtime unit tests.

### DIFF
--- a/src/components/Dialog/Dialog.spec.jsx
+++ b/src/components/Dialog/Dialog.spec.jsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import { mount, shallow } from 'enzyme';
-import { common } from '../../util/generic-tests';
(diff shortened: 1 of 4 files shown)


dsusskind:
I like it! Although why not keep our `FC` type as an interface with the properties we want, but just don't extend from `React.FC`?

jondlm:
That would certainly be ideal but I couldn't find a way to make that work. If you want to take a stab at it I'd welcome it but my hunch is that it's not possible with the way Typescript currently works. Something about the way that React does special magic with the tacked on `defaultProps` gets broken every time I try to assign any concrete type to the component.

dsusskind:
Oh I see what you mean. Looks like this is the best we can do then.

jondlm:
Closing this PR as I'm doing this work on #1058

PR #1075, opened 2019-11-19
[proposal] enforce propType descriptions and allow passing a Generic to `createClass`

jhsu:
benefits:
- `this.props` hints in the render function
- enforce that `propTypes` on the definition has prop-type descriptions

note: tests will break with just this change

Screen Shot 2019-11-19 at 11 01 49 AM

when `propTypes` is missing something from the props interface:
Screen Shot 2019-11-19 at 11 00 23 AM

### DIFF
--- a/src/util/component-types.ts
+++ b/src/util/component-types.ts
@@ -1,4 +1,4 @@
-import React, { ReactElement } from 'react';
+import React from 'react';
 import createReactClass from 'create-react-class';
 import PropTypes from 'react-peek/prop-types';
 import _ from 'lodash';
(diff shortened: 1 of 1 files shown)


dsusskind:
Nice, thanks!

jondlm:
Looks like the compiler isn't happy about some stuff: https://travis-ci.org/appnexus/lucid/jobs/614161934

I can take a look at it later if you don't have the bandwidth.

jhsu:
@jondlm fixed (chromatic still errors)

jondlm:
Awesome, yeah looks like chromatic doesn't like running on forks. I'll merge this.

PR #1076, opened 2019-11-19
Accordion typescript conversion

dreidn:
## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [x] Edge (Win 10)

Docspot Lucid Docs:
https://docspot.adnxs.net/projects/lucid/accordian-typescript-conversion/?path=/docs/layout-accordion--basic


src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,141 @@
...
+			event,
+			props: this.props,
+		});
+	}
+
+	render() {
+		const { style, className, selectedIndex, ...passThroughs } = this.props;
+
+		const itemChildProps = _.map(
+			findTypes(this.props, Accordion.Item as { propName?: string | string[] }),
```
dreidn: I'm not sure why this needs to be type asserted. Just passing `Accordion.Item` doesn't type check. A lot of components use `findTypes` without the type assertion so I would assume it should work.

dreidn: error for reference
```
Argument of type 'Readonly<Overwrite<DetailedHTMLProps<HTMLAttributes<HTMLDivElement>, HTMLDivElement>, IAccordionPropsRaw>> & Readonly<{ children?: ReactNode; }>' is not assignable to parameter of type 'IExpanderPanelProps'.
```

src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,141 @@
...
+	 * Prop alternative to Header child component passed through to the
+	 * underlying ExpanderPanel
+	 */
+	Header: IExpanderPanelHeaderProps;
+}
+
+type IAccordionProps = Overwrite<
+	React.DetailedHTMLProps<React.HTMLAttributes<HTMLDivElement>, HTMLDivElement>,
+	IAccordionPropsRaw
+>;
```
dsusskind: Do we need to use Overwrite here? Or can we just name the above interface `IAccordionProps`?

dreidn: I used Overwrite because `onSelect` in `IAccordionPropsRaw` conflicts with `React.DetailedHTMLProps...`. It seems like we aren't even using `React.DetailedHTMLProps...` in other class components, so I'm guessing I don't actually need to extend it and can just use `IAccordionPropsRaw` as ``IAccordionProps` like you said.

src/components/Accordion/Accordion.tsx
dsusskind: Thinking about this some more, I think maybe this pattern should not be using ExpanderPanel directly. Like how CalendarMonth gets used here: https://github.com/appnexus/lucid/blob/5ab4bb2fb861fd38c92e12e07f46d1fbbdef735f/src/components/DateSelect/DateSelect.tsx.

So Accordian.Item should not actually be an ExpanderPanel and should just be some dummy like
```
const AccordianItem = (_props: IAccordianItemProps): null =>
	null;
```

What do you think?

src/components/Accordion/Accordion.tsx
dsusskind: Oh yea, I think you are right to use it then. Otherwise we'll get some browser errors due to conflicting signatures for onSelect. Sorry for the confusion.

src/components/Accordion/Accordion.reducers.ts
```diff
@@ -0,0 +1,8 @@
+import { IAccordionState } from './Accordion';
+
+export function onSelect(state: IAccordionState, selectedIndex: number) {
```
jondlm: Maybe put an explicit return type of `IAccordionState` here as well.

src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,141 @@
+import _ from 'lodash';
+import React from 'react';
+import PropTypes from 'react-peek/prop-types';
+import { lucidClassNames } from '../../util/style-helpers';
+import { buildHybridComponent } from '../../util/state-management';
```
jondlm: Let's adjust this to use `buildModernHybridComponent` instead. See ToolTip for an example of how to use it.

src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,141 @@
...
+type IAccordionProps = Overwrite<
+	React.DetailedHTMLProps<React.HTMLAttributes<HTMLDivElement>, HTMLDivElement>,
+	IAccordionPropsRaw
+>;
+
+export interface IAccordionState {
+	selectedIndex: number | null;
+}
+
+const defaultProps = {
```
jondlm: So, looking at these default props when combined with `IAccordionPropsRaw` that means that `selectedIndex` and `Header` will be required but I don't think that's correct. There shouldn't be any required props.

src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,141 @@
...
+		`,
+		style: object`
+			Passed through to the root element.
+		`,
+		Header: any`
+			Prop alternative to Header child component passed through to the
+			underlying ExpanderPanel
+		`,
+	};
+	static definition = {
```
jondlm: Shouldn't need this `definition` object anymore with `buildModernHybridComponent`. However you will want to stick the `peek` property on as a class static.

src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,141 @@
...
+		selectedIndex: number`
+			Indicates which item is expanded
+		`,
+		onSelect: func`
+			Called when the user clicks on the component's header of an item.
+		`,
+		style: object`
+			Passed through to the root element.
+		`,
+		Header: any`
```
jondlm: After reviewing the component  I think that is actually wrong at the `Accordion` level. The header is actually at the child level if you look through the examples. I think we can just remove this one from here as well from the type signature.

src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,141 @@
...
+
+const cx = lucidClassNames.bind('&-Accordion');
+
+const { any, func, object, number, string } = PropTypes;
+
+interface IAccordionPropsRaw extends StandardProps {
+	/**
+	 * Indicates which item is expanded
+	 * */
+	selectedIndex: number;
```
jondlm: I think this is optional and also allows `null` if I'm reading the code right

src/components/Accordion/Accordion.tsx
jondlm: I think you're right that using a dummy child component would work. I'm not sure it's necessary as long as we're getting the right type signatures for the consumers. I.e. might not be worth the effort/risk if this is working okay with a type cast.

src/components/Accordion/Accordion.tsx
dreidn: I think this may have been reviewed before I updated it to use `buildModernHybridComponent`, but it does use it now

src/components/Accordion/Accordion.tsx
dreidn: Ah good to know. I converted it to a functional component when I added `buildModernHybridComponent`,  so it has the right static property `peek` now.

src/components/Accordion/Accordion.tsx
dreidn: You're right, good catch. I noticed this when I converted examples to typescript, they did not typecheck.

src/components/Accordion/Accordion.tsx
dreidn: You're right, looking at the code it does look like it should allow `null`. I'll change this to `number | null`

src/components/Accordion/Accordion.tsx
dreidn: That is a good point. It doesn't seem to even work if you pass a header as a prop at the `Accordion` level. However, removing the prop causes a common test to fail. Seems like this test checks for child components defined as static properties, and makes sure a prop exists for them:
https://github.com/appnexus/lucid/blob/ebe298ecf66b87457fa2665a0de2dd3cbcaea39f/src/util/generic-tests.jsx#L136

jondlm: Looks like you can pass an `exemptChildComponents` prop to the common tests to bypass that.

src/components/Accordion/Accordion.reducers.spec.js
```diff
@@ -1,5 +1,5 @@
 import assert from 'assert';
-import * as reducers from './Accordion.reducers.js';
+import * as reducers from './Accordion.reducers.ts';
```
yasskin: I think you can leave off the file extension type here, so just `'./Accordion.reducers';

src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,139 @@
...
+	Header: any`
+		Prop alternative to Header child component passed through to the
+		underlying ExpanderPanel
+	`,
+};
+
+Accordion.peek = {
+	description: `
+		This is a container that renders panels and controls its
+		expansion/retraction.
```
yasskin: Not an engineering issue, but the description will read better if we start by stating the component name instead of an ambiguous "This", so maybe:
\`Accordion\` is a container that renders multiple panels and controls their expansion or collapse."

jondlm:
Let's merge master into this branch so chromatic can match the story snapshots up correctly.

PR #1080, opened 2019-11-21
Functional component fix

jondlm:


src/components/Icon/AddURLIcon/AddURLIcon.tsx
```diff
@@ -2,16 +2,16 @@ import _ from 'lodash';
...
 const cx = lucidClassNames.bind('&-AddURLIcon');
 
 interface IAddURLIconProps extends IIconProps {}
 
-export const AddURLIcon: FC<IAddURLIconProps> = ({
+export const AddURLIcon = ({
 	className,
 	...passThroughs
-}): React.ReactElement => {
+}: IAddURLIconProps) => {
```
dsusskind: is the `React.ReactElement` return type not needed for Icons?

jondlm: There were some cases that made it weird to add an explicit return type so I called an audible and removed the eslint rule that required them. It might be over reaching now that I'm thinking about it today.

jondlm: I went ahead and restored the explicit return types to all the icons.

PR #1087, opened 2019-12-10
Improve VerticalListMenu types and make createClass more permissive

jondlm:
## PR Checklist

  - [ ] Chrome
  - [x] Firefox
  - [ ] Safari
  - [ ] Edge (Win 10)


src/util/component-types.ts
```diff
@@ -48,7 +48,8 @@ type TypesType<P> =
 	| { propName?: string | string[] }
 	| any; // TODO: figure out a type that works with inferred functional components
 
-interface ICreateClassComponentSpec<P, S> extends React.Mixin<P, S> {
```
jondlm: @jhsu I added the `any` default here because the type inference was doing really odd things for folks who didn't explicitly pass types to the generic.

src/util/component-types.ts
```diff
@@ -63,7 +64,11 @@ interface ICreateClassComponentSpec<P, S> extends React.Mixin<P, S> {
 	// TODO: improve these with a stricter type https://stackoverflow.com/a/54775885/895558
 	reducers?: { [K in keyof P]?: (arg0: S, ...args: any[]) => S };
 	selectors?: { [K in keyof P]?: (arg0: S) => any };
-	render?(this: { props: P }): React.ReactNode;
+	render?(
+		this: { props: P } & {
+			[k: string]: any;
```
jondlm: @jhsu this was needed cause folks can drop additional properties on the class definition that would show up on `this` it turns out.

jhsu: oh interesting, didn't know that

src/util/component-types.ts
jondlm: Actually turns out adding any types as defaults still allows the TS compiler to infer the types. Users will still need to manually add `<any, any>` to bypass most of the checking. I removed the defaults after more testing.

PR #1089, opened 2019-12-13
Cxp 805 validation state to search select fields

trippygif:
## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [ ] Edge (Win 10)

// please check out SearchableMultiSelect, SearchableSingleSelect, SearchableSelect
https://docspot.adnxs.net/projects/lucid/CXP-805-validation-state-to-search-select-fields/?path=/docs/controls-searchablemultiselect--invalid


trippygif:
updated test cases for more coverage

src/components/SearchableMultiSelect/SearchableMultiSelect.jsx
```diff
@@ -547,14 +561,15 @@ const SearchableMultiSelect = createClass({
 			<div
 				{...omitProps(passThroughs, SearchableMultiSelect)}
 				className={cx('&', className)}
+				Error={errorChildProps}
```
jondlm: Let's leave this out since `div` won't accept this prop and will probably throw a React warning about it in the console.

src/components/SearchableMultiSelect/SearchableMultiSelect.jsx
```diff
@@ -533,6 +544,9 @@ const SearchableMultiSelect = createClass({
 			'props',
 			{}
 		);
+		const errorChildProps = _.first(
+			_.map(findTypes(props, Validation.Error), 'props')
```
jondlm: Nice reuse of the component!

src/components/SearchableMultiSelect/SearchableMultiSelect.jsx
```diff
@@ -580,6 +595,7 @@ const SearchableMultiSelect = createClass({
 								'&-search',
 								{
 									'&-search-is-small': isSmall,
+									'&-search-is-error': errorChildProps && errorChildProps.children && errorChildProps.children !== true
```
jondlm: I'm a little confused by the logic here. Suppose I do pass `Error={true}`, wouldn't that result in this expression being `false`?

src/components/SearchableMultiSelect/SearchableMultiSelect.spec.jsx
```diff
@@ -233,6 +233,34 @@ describe('SearchableMultiSelect', () => {
...
+						<Option>option b</Option>
+					</SearchableMultiSelect>
+				);
+
+				const searchWrapper = wrapper.find('.lucid-SearchableMultiSelect-search-is-error');
+
+				expect(searchWrapper).toBeTruthy();
+			});
+
+			it('should render out the error div', () => {
```
jondlm: Maybe add a test case specifically for the `true` case that should render the input in an `-is-error` state but _not_ render the extra div

src/components/SearchableSelect/SearchableSelect.jsx
```diff
@@ -367,12 +378,12 @@ const SearchableSelect = createClass({
 		return visibleOptionsCount > 0 ? (
 			options
 		) : (
-			<DropMenu.Option isDisabled>
-				<span className={cx('&-noresults')}>
-					No results match "{searchText}"
+				<DropMenu.Option isDisabled>
```
jondlm: We are using prettier which should handle all this formatting stuff. I'd suggest setting up your editor to run the code through prettier on save

src/components/SearchableSelect/SearchableSelect.jsx
```diff
@@ -405,88 +416,109 @@ const SearchableSelect = createClass({
...
+		);
+		
+
 		const placeholder = _.get(placeholderProps, 'children', 'Select');
 		const isItemSelected = _.isNumber(selectedIndex);
 
 		return (
-			<DropMenu
-				{...dropMenuProps}
+			<div Error={errorChildProps}
```
jondlm: Doesn't make sense to pass `Error` to `div` here either.

src/components/SearchableSelect/SearchableSelect.jsx
```diff
@@ -405,88 +416,109 @@ const SearchableSelect = createClass({
...
-							'&-Control-is-disabled': isDisabled,
-						})}
-					>
-						<span
-							{...(!isItemSelected ? placeholderProps : null)}
-							className={cx(
-								'&-Control-content',
-								!isItemSelected ? _.get(placeholderProps, 'className') : null
-							)}
+				<DropMenu
```
jondlm: It's a little tough to see the diff because of formatting stuff. Maybe run prettier on this and then it'll be more clear?

src/components/SearchableSelect/SearchableSelect.jsx
trippygif: will get the prettier set up and update the code

src/components/SearchableMultiSelect/SearchableMultiSelect.jsx
trippygif: where would you suggest moving the error props? I wrapped everything in a div so I could separately render out the error text underneath the DropMenu component

src/components/SearchableMultiSelect/SearchableMultiSelect.jsx
trippygif: good find, I re-used some of the logic for the error-content div logic and forgot to update it for the case of no text. got it updated now

jondlm: I don't think we need to do anything with them. The only thing we care about is the `children` which you're already using. I just took a look at `Validation` and it claims to have a `description` prop but I think that's wrong. The only thing it seems to care about is `children`.

src/components/SearchableMultiSelect/SearchableMultiSelect.spec.jsx
trippygif: updated/added test case with the revised search class logic

src/components/SearchableMultiSelect/SearchableMultiSelect.jsx
trippygif: right, the error border should be applied if there are children but if the child is true then we know there's a general validation applied to the component. thus removing that last piece of logic will allow passing `Error={true}`  to validate the search field but not have a message. I added a test case to check this, will push up shortly :)

src/components/SearchableMultiSelect/SearchableMultiSelect.jsx
```diff
@@ -549,12 +563,13 @@ const SearchableMultiSelect = createClass({
 				className={cx('&', className)}
 			>
 				<DropMenu
+					Error={errorChildProps}
```
dreidn: Does it make sense to pass an `Error` prop here? It doesn't seem like DropMenu takes an `Error` prop.

trippygif: I wasn't entirely sure where to pass in the error props; I initially wrapped the entire component in a div and passed the error props in that container but Jon indicated that it would cause some React errors in the console. I was unsure where to pass the `Error` prop outside of the div since the other components do not take an Error prop.

trippygif: @dreidn I am realizing that the `Error` prop is not necessary in any of the components because they are not expecting it. I am removing passing the `Error` prop and simply relying on the errorChildProps driving the validation logic in all three of the components. Forgive my mistakes :)

src/components/SearchableSingleSelect/SearchableSingleSelect.jsx
```diff
@@ -402,9 +415,22 @@ const SearchableSingleSelect = createClass({
 								? this.renderOptionContent(selectedOptionProps, '')
 								: selectedOptionProps.Selected
 						}
+						className={cx('&', {
+							'&-select-error': errorChildProps && errorChildProps.children && errorChildProps.children !== true
```
jondlm: I think this is the same logic issue as above that you fixed elsewhere. Also, should it be `&-select-is-error` instead?

jondlm:
I also noticed the spacing on the validation message doesn't look quite right. It should be closer to the input than it is.

src/components/SearchableSingleSelect/SearchableSingleSelect.jsx
trippygif: @jondlm my most recent changes should have this updated

trippygif:
https://docspot.adnxs.net/projects/lucid/CXP-805-validation-state-to-search-select-fields/?path=/docs/*       --> latest changes

PR #1092, opened 2019-12-19
TypeScript convert paginator (869) and singleselect (877)

yasskin:
## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari

Storybook Paginator can be viewed [here](https://docspot.adnxs.net/projects/lucid/ts-convert-paginator-singleselect/?path=/docs/navigation-paginator--basic)

Storybook SingleSelect can be viewed [here](https://docspot.adnxs.net/projects/lucid/ts-convert-paginator-singleselect/?path=/docs/false-singleselect--basic)


src/components/DropMenu/DropMenu.tsx
```diff
@@ -552,8 +554,15 @@ class DropMenu extends React.Component<IDropMenuProps, IDropMenuState> {
 	};
 
 	static preprocessOptionData = (
-		props: IDropMenuProps,
-		ParentType = DropMenu
+		props: StandardProps,
+		//TODO: ask about this fix
+		ParentType: {
```
dreidn: This needed to support passing components other than DropMenu as ParentType. After taking a look at it, I thought `typeof` seemed a bit cleaner than using generic types since it would require 4 type arguments.

dreidn: whether we keep this type or create a new one we should remove the TODO comment

src/components/DropMenu/DropMenu.tsx
```diff
@@ -552,8 +554,15 @@ class DropMenu extends React.Component<IDropMenuProps, IDropMenuState> {
...
+			OptionGroup: typeof OptionGroup;
+			Option: typeof Option;
+			NullOption: typeof NullOption;
+			FixedOption: typeof FixedOption;
+		}
+		// ParentType: typeof DropMenu | typeof SingleSelect | typeof = DropMenu
```
dreidn: remove comment

src/components/Paginator/Paginator.reducers.ts
```diff
@@ -0,0 +1,57 @@
...
+		SingleSelect: SingleSelect.onSelect(
+			state.SingleSelect,
+			selectedPageSizeIndex
+		),
+	};
+}
+
+export { SingleSelect, TextField };
+
+// import _ from 'lodash';
```
dreidn: remove comments

src/components/Paginator/Paginator.selectors.ts
```diff
@@ -0,0 +1,44 @@
...
+		? 1
+		: Math.floor(totalCount / pageSize) + _.clamp(totalCount % pageSize, 0, 1);
+}
+
+export default {
+	totalPages: totalPagesSelector,
+};
+
+// import _ from 'lodash';
+
```
dreidn: remove comments

src/components/TextField/TextField.reducers.ts
```diff
@@ -4,14 +4,26 @@ export interface ITextFieldState {
...
+): ITextFieldState {
+	return {
+		...state,
+		value: value,
+	};
+}
+
+export default { onChange };
+
+// export default {
```
dreidn: remove comment

src/components/DropMenu/DropMenu.tsx
yasskin: thx. removed.

src/components/DropMenu/DropMenu.tsx
yasskin: thx. removed.

src/components/Paginator/Paginator.reducers.ts
yasskin: thx. removed.

src/components/TextField/TextField.reducers.ts
yasskin: thx. removed.

src/components/DropMenu/DropMenu.tsx
```diff
@@ -108,14 +111,14 @@ OptionGroup.defaultProps = {
...
-interface IDropMenuOptionProps extends StandardProps {
-	description?: string;
+export interface IDropMenuOptionProps extends StandardProps {
 	isDisabled?: boolean;
 	isHidden?: boolean;
 	isWrapped?: boolean;
 }
 
-const Option = (_props: IDropMenuOptionProps): null => null;
+const Option: FC<IDropMenuOptionProps> = (_props: IDropMenuOptionProps): null =>
```
jondlm: I might be reading between the lines but it seems like you needed to stop relying on type inference because you needed the types elsewhere. A couple questions if that's the case:

1. would simply using `typeof` work better instead of `FC`? The only issues I could see with the explicit `FC` type would be related to default props but I don't think any of these child components are using default props
1. if we do stick with the explicit `FC` type then I think we can remove the `_props: IDropMenuOptionProps` for example since that was only there to aid type inference

src/components/Paginator/Paginator.reducers.spec.js
```diff
@@ -1,6 +1,6 @@
 import assert from 'assert';
 import * as SingleSelect from '../SingleSelect/SingleSelect.reducers';
-import { onPageSelect, onPageSizeSelect } from './Paginator.reducers.js';
+import { onPageSelect, onPageSizeSelect } from './Paginator.reducers';
```
jondlm: Good catch

src/components/Paginator/Paginator.tsx
```diff
@@ -0,0 +1,340 @@
...
+	totalPages: undefined,
+	pageSizeOptions: [10, 50, 100],
+	SingleSelect: {
+		...SingleSelect.defaultProps,
+		selectedIndex: 0,
+	},
+	TextField: TextField.defaultProps,
+};
+
+class Paginator extends React.Component<IPaginatorProps, IPaginatorState> {
```
jondlm: FWIW I think this component could have been converted to a functional component since it doesn't appear to be using any lifecycle methods.

src/components/SingleSelect/SingleSelect.tsx
```diff
@@ -22,96 +36,179 @@ const {
...
+			props,
+			event,
+		}: {
+			props: ISingleSelectOptionProps | undefined;
+			event: React.MouseEvent | React.KeyboardEvent;
+		}
+	) => void;
+
+	/** TODO: doublecheck this type */
+	ref?: string;
```
jondlm: This does indeed seem weird. Was it needed for some reason?

src/components/SingleSelect/SingleSelect.tsx
```diff
@@ -254,7 +360,7 @@ const SingleSelect = createClass({
 			<DropMenu
 				{...dropMenuProps}
 				isDisabled={isDisabled}
-				selectedIndices={isItemSelected ? [selectedIndex] : []}
+				selectedIndices={isItemSelected ? [selectedIndex as number] : []}
```
jondlm: I double checked and this seems fine since it's already been run through `isNumber`

src/util/state-management.ts
```diff
@@ -93,6 +107,7 @@ export function bindReducersToState<P, S extends object>(
 	reducers: Reducers<P, S>,
 	{ getState, setState }: IStateOptions<S>
 ) {
+	console.log('getDeepPaths', getDeepPaths(reducers));
```
jondlm: Looks like a lonely, forgotten console log

src/util/state-management.ts
```diff
@@ -106,12 +121,17 @@ export function bindReducersToState<P, S extends object>(
...
+/*
+
+*/
 export function getStatefulPropsContext<P, S extends object>(
 	reducers: Reducers<P, S>,
 	{ getState, setState }: IStateOptions<S>
 ): IBoundContext<P, S> {
 	const boundReducers = bindReducersToState(reducers, { getState, setState });
 
+	console.log('boundReducers', boundReducers);
```
jondlm: Another one

jondlm:
Sorry it took me so long to review this one. It's a doozy :D

src/components/DropMenu/DropMenu.tsx
dreidn: I think this may just be a remnant of us trying to debug a type error involving the Option child component. I removed all the uses of FC in DropMenu and it looks like we can get away without it.

src/components/SingleSelect/SingleSelect.tsx
dreidn: This was a tricky one.

Since we were using `Overwrite` for the `ISingleSelectPropsRaw` interface, typescript seemed to think we wanted a `div` ref prop, not a `SingleSelect` ref, causing this line to give a type error: https://github.com/appnexus/lucid/blob/65d922390e79d8f1fc5320e368a4bd1b710ae472/src/components/Paginator/Paginator.jsx#L180

Looking at it again, I don't think we should've been using Overwrite for SingleSelect in the first place since `SingleSelect` doesn't pass through props to any native HTML tags. I removed it, and now everything typechecks without declaring `ref` in `ISingleSelectProps`.

src/components/SingleSelect/SingleSelect.tsx
dreidn: I get an error if I remove the `as number`, typescript doesn't seem to pick up the `isNumber` check through `isItemSelected`.

PR #1104, opened 2020-01-07
Tag border colors

gschnall:
Changes requested for Shedules on buyside

**Updated** - Examples now show the following:
Screen Shot 2020-01-07 at 10 35 53 AM

removable tags keep their primary color border by default

## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [ ] Edge (Win 10)

### DIFF
--- a/src/components/Sidebar/__snapshots__/Sidebar.spec.jsx.snap
+++ b/src/components/Sidebar/__snapshots__/Sidebar.spec.jsx.snap
@@ -2342,6 +2342,7 @@ exports[`Sidebar [common] example testing should match snapshot(s) for 11.differ
           isRemovable={true}
           isTop={false}
           key="0"
+          kind="default"
           onRemove={[Function]}
         >
           Tag # 0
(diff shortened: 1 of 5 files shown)


TylerJBrown192:
@rsadres You had Gabe remove the other color examples but leave the code in for the other `kind` variations. Do you want to remove the possible variations from the props table as well?

Screen Shot 2020-01-07 at 11 47 50 AM

rsadres:
> @rsadres You had Gabe remove the other color examples but leave the code in for the other `kind` variations. Do you want to remove the possible variations from the props table as well?
> 
> Screen Shot 2020-01-07 at 11 47 50 AM

I don't think so Tyler, but thanks for checking. I might have thought about not having that option before it was built. I don't think we need to remove code. I'm ok with guiding desired behavior rather than enforcing it. I'm sure someone will think of a use for the other options in the future.

gschnall:
@TylerJBrown192 Good catch! I considered changing the listed variations but I figured it made sense to have it reflect what's currently in the codebase.

PR #1110, opened 2020-01-10
Avoid adding kind border styles on parent tags

gschnall:
Avoid activating kind props on parent tags by using a recursive function to set the kind border color on leaves.

- Also removed the kind type of primary; since it isn't being assigned a style/used.

## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [ ] Edge (Win 10)

### DIFF
--- a/src/components/Tag/Tag.tsx
+++ b/src/components/Tag/Tag.tsx
@@ -34,7 +34,7 @@ export interface ITagProps
 	isRemovable: boolean;
 
 	/** Style variations of the `Tag`. */
-	kind?: 'primary' | 'success' | 'warning' | 'danger' | 'info' | 'default';
+	kind?: 'success' | 'warning' | 'danger' | 'info' | 'default';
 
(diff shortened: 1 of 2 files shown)


src/components/Tag/Tag.tsx
```diff
@@ -34,7 +34,7 @@ export interface ITagProps
 	isRemovable: boolean;
 
 	/** Style variations of the `Tag`. */
-	kind?: 'primary' | 'success' | 'warning' | 'danger' | 'info' | 'default';
```
jondlm: Isn't this technically a breaking change?

gschnall: We never included a color for primary in the css. It's the same color as info. I can keep it there if you'd like and give it the same color blue. Would you prefer that?

gschnall: @jondlm I've added primary back as an option as you suggested and gave it style in the css.

PR #1130, opened 2020-02-13
ResponsiveGrid component

dreidn:
Added `ResponsiveGrid` component to handle the use case of displaying a grid of cards within a container that might resize. The grid will responsively display one, two, or three columns depending on the width of the container.  Added an example of using `ResponsiveGrid` within the `Sidepanel` component.

## PR Checklist

Storybook can be viewed [here](https://docspot.adnxs.net/projects/lucid/sidepanel-with-grids/?path=/story/utility-responsivegrid--basic)

  - [x] Chrome
  - [x] Firefox
  - [x] Safari


src/components/ResponsiveGrid/ResponsiveGrid.tsx
```diff
@@ -0,0 +1,187 @@
...
+
+const { string, number, arrayOf } = PropTypes;
+
+interface IResponsiveGridCellProps extends StandardProps {}
+
+const Cell = (_props: IResponsiveGridCellProps): null => null;
+Cell.displayName = 'ResponsiveGrid.Cell';
+Cell.peek = {
+	description: `
+		Renders a \`<span>\` as the grid cell
```
TylerJBrown192: Should this say `<article>`, per line 77 in this file?

src/components/ResponsiveGrid/ResponsiveGrid.tsx
```diff
@@ -0,0 +1,187 @@
...
+		const columns = _.reduce(
+			/* We use ts-ignore below since the typescript compiler pattern matches this _.map
+			 to an array of booleans */
+			_.map(cellProps, (props: IResponsiveGridCellProps, key: number) => (
+				<article key={key} className={cx('&-Cell')}>
+					{props.children}
+				</article>
+			)),
+			(columns: React.ReactElement[][], cell, idx: number) => {
+				//@ts-ignore
```
TylerJBrown192: This issue (and the above comment regarding `_.map`'s return type being `boolean[]`) can be resolved by declaring cellProps with its Interface type on line 68, e.g. `const cellProps: IResponsiveGridCellProps[]`

src/components/ResponsiveGrid/ResponsiveGrid.tsx
```diff
@@ -0,0 +1,187 @@
...
+
+		return (
+			<div
+				className={cx('&', {
+					'&-one-column': numberOfColumns === 1,
+				})}
+			>
+				{_.map(columns, (col, key) => {
+					return (
+						<div key={key} className={cx('&-Column')}>
```
TylerJBrown192: Should we declare the `key` as something other than the index of the column, since this has the possibility of being re-rendered if a breakpoint changes? A better `key` value isn't immediately popping out at me though.

src/components/ResponsiveGrid/ResponsiveGrid.tsx
```diff
@@ -0,0 +1,187 @@
...
+	findTypes,
+	omitProps,
+} from '../../util/component-types';
+import Resizer from '../Resizer/Resizer';
+
+const cx = lucidClassNames.bind('&-ResponsiveGrid');
+
+const { string, number, arrayOf } = PropTypes;
+
+interface IResponsiveGridCellProps extends StandardProps {}
```
yasskin: interface IResponsiveGridCellProps extends StandardProps {};

src/components/ResponsiveGrid/ResponsiveGrid.tsx
dreidn: yep, good catch!

src/components/ResponsiveGrid/ResponsiveGrid.tsx
dreidn: Yeah, I didn't see any options for key here other than the index. Although different groups of columns are rendered as the width changes, the ordering of columns within a group never changes once rendered, so I think it is safe to use the index as a key here.
https://reactjs.org/docs/lists-and-keys.html

TylerJBrown192: Ah right, it's re-ordering that is the problematic case, not re-rendering the same list. Agreed on the index issue, thank you!

PR #1136, opened 2020-03-02
TXP-127 add vertical orientation to dots icon

danvanhorn:
## PR Checklist

Add vertical option to the three dots icon.

Storybook can be viewed [here](https://docspot.adnxs.net/projects/lucid/txp-127-add-vertical-orientation-to-dots-icon/?path=/story/icons-icons-dotsicon--default)

  - [x] Chrome
  - [x] Firefox
  - [x] Safari

### DIFF
--- a/src/components/Icon/DotsIcon/DotsIcon.tsx
+++ b/src/components/Icon/DotsIcon/DotsIcon.tsx
@@ -61,7 +61,7 @@ DotsIcon.peek = {
 DotsIcon.propTypes = {
 	...iconPropTypes,
 	direction: oneOf(_.values(Orientation))`
-		Sets the orientation of how the dots are displayed. 
+		Sets the orientation of how the dots are displayed. Defaults to 'horizontal'. 
 	`,
 };
 
--- a/src/components/Icon/DotsIcon/DotsIcon.tsx
+++ b/src/components/Icon/DotsIcon/DotsIcon.tsx
@@ -65,4 +65,9 @@ DotsIcon.propTypes = {
 	`,
 };
 
+DotsIcon.defaultProps = {
+	...Icon.defaultProps,
+	direction: Orientation.horizontal,
+};
+
 export default DotsIcon;
--- a/src/components/Icon/DotsIcon/DotsIcon.tsx
+++ b/src/components/Icon/DotsIcon/DotsIcon.tsx
@@ -9,22 +9,17 @@ import PropTypes from 'react-peek/prop-types';
 const cx = lucidClassNames.bind('&-DotsIcon');
 
 const { oneOf } = PropTypes;
-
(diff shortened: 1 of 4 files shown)


src/components/Icon/DotsIcon/DotsIcon.tsx
```diff
@@ -4,16 +4,36 @@ import Icon, { IIconProps, propTypes as iconPropTypes } from '../Icon';
...
 
-interface IDotsIconProps extends IIconProps {}
+const { oneOf } = PropTypes;
+
+enum Orientation {
+	vertical = 'vertical',
+	horizontal = 'horizontal'
+}
+interface IDotsIconProps extends IIconProps {
+	direction?: keyof typeof Orientation;
```
jhsu: instead of using an enum, i'd suggest just using string literals `'vertical' | 'horizontal'` based on how you are using it here.

if you expect consumers to import your enum type and pass a value for a prop, you could use `direction?: Orientation`, but I don't think that's what you want.

src/components/Icon/DotsIcon/DotsIcon.tsx
```diff
@@ -4,16 +4,36 @@ import Icon, { IIconProps, propTypes as iconPropTypes } from '../Icon';
...
+}
 
 export const DotsIcon = ({
 	className,
+	direction = Orientation.horizontal,
 	color = Color.primary,
 	...passThroughs
 }: IDotsIconProps) => {
+	const isVerticalOrientation = _.isEqual(direction, Orientation.vertical);
```
jhsu: `_.isEqual` is good for deep comparisons, but here you're just doing a string comparison, so it's probably simpler to do a `===` compare.

src/components/Icon/DotsIcon/DotsIcon.tsx
danvanhorn: Either way is fine with me. I've seen it done with both string literals and with enums in this codebase.

src/components/Icon/DotsIcon/DotsIcon.tsx
danvanhorn: done

src/components/Icon/DotsIcon/DotsIcon.tsx
yasskin: `Legend` has an example of the string literal. I see that we use enums all over the place too.

danvanhorn: @jhsu @yasskin would you prefer then that we stick with string literals? I just changed it back, or rather consolidate both string literals into an "orientation enum" and use that in both places. I personally think using the enum is superior to using string literals, but this is your repo so it's your call.

PR #1138, opened 2020-03-05
update min-width for table

johnclayshi:
## PR Checklist

  - [x] Chrome
  - [x] Firefox
  - [x] Safari


src/components/Table/Table.tsx
```diff
@@ -252,6 +252,9 @@ interface IThProps
 	/** Sets the width of the cell. */
 	width?: number | string;
 
+	/** Sets the min width of the cell. */
+	minWidth?: number | string;
```
danvanhorn: Are there any times where we need to convert `minWidth` from a string to number. Similar to what is being done on line 498.

src/components/DataTable/__snapshots__/DataTable.spec.jsx.snap
```diff
@@ -20210,4 +20210,4 @@ exports[`DataTable [common] example testing should match snapshot(s) for 14.resi
     </div>
   </div>
 </EmptyStateWrapper>
-`;
+`;
```
danvanhorn: can we fix this?

PR #1150, opened 2020-05-15
Deal 1429 side panel improvement

pearlanre:
## PR Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari

### DIFF
--- a/src/components/SidePanel/SidePanel.tsx
+++ b/src/components/SidePanel/SidePanel.tsx
@@ -134,6 +134,12 @@ class SidePanel extends React.Component<ISidePanelProps, ISidePanelState, {}> {
 			Sets the initial width in pixels. The actual width may change if the user
 			resizes it.
 		`,
+		minWidth: number`
+		Sets the minimum width of the Side Panel.
+		`,
+		maxWidth: number`
+		Sets the maximum width of the Side Panel.
+		`,
 		topOffset: oneOfType([number, string])`
 			Sets the top margin for the panel. Defaults to \`0\`.
 		`,
@@ -156,6 +162,8 @@ class SidePanel extends React.Component<ISidePanelProps, ISidePanelState, {}> {
 		preventBodyScroll: false,
 		topOffset: 0,
 		width: 240,
+		minWidth: 200,
+		maxWidth: 500,
 	};
 
 	static Header = SidePanelHeader;
@@ -181,10 +189,11 @@ class SidePanel extends React.Component<ISidePanelProps, ISidePanelState, {}> {
 		{ dX }: { dX: number },
 		{ event }: { event: MouseEvent | TouchEvent }
 	): void => {
-		const { startWidth } = this.state;
+		const { startWidth, width } = this.state;
 		this.setState({
 			width: startWidth + dX * (this.props.position === 'right' ? -1 : 1),
 			isResizing: false,
+			startWidth:width,
 		});
 		this.props.onResize(startWidth - dX, { props: this.props, event });
 	};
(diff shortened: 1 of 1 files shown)


src/components/SidePanel/SidePanel.tsx
```diff
@@ -134,6 +141,12 @@ class SidePanel extends React.Component<ISidePanelProps, ISidePanelState, {}> {
 			Sets the initial width in pixels. The actual width may change if the user
 			resizes it.
 		`,
+		minWidth: number`
+		Sets the minimum width of the Side Panel.
+		`,
+		maxWidth: number`
```
jondlm, line 147: As I'm looking at this I'm realizing what would be _really_ nice if we could prevent the panel from going off the left side of the screen. Would it be possible to support something like `100%`? I haven't dug into this component so I don't even know if it's technically feasible.

src/components/SidePanel/SidePanel.tsx
```diff
@@ -156,6 +169,8 @@ class SidePanel extends React.Component<ISidePanelProps, ISidePanelState, {}> {
 		preventBodyScroll: false,
 		topOffset: 0,
 		width: 240,
+		minWidth: null,
```
jondlm, line 172: Might make your logic easier if you used `-Infinity` and `Infinity`

src/components/SidePanel/SidePanel.tsx
```diff
@@ -172,19 +187,23 @@ class SidePanel extends React.Component<ISidePanelProps, ISidePanelState, {}> {
 
 	handleResize = ({ dX }: { dX: number }): void => {
 		const { startWidth } = this.state;
+		const position = startWidth + dX * (this.props.position === 'right' ? -1 : 1);
+		const adjustedWidth = !_.isNil(this.props.minWidth) && position < this.props.minWidth ? this.props.minWidth
```
jondlm, line 191: Take a look at https://lodash.com/docs/4.17.15#clamp and see if it would simplify this.

pearlanre: ah thanks. I thought there was a lodash implementation for this but couldn't find what it was called.

src/components/SidePanel/SidePanel.tsx
```diff
@@ -156,6 +168,8 @@ class SidePanel extends React.Component<ISidePanelProps, ISidePanelState, {}> {
 		preventBodyScroll: false,
 		topOffset: 0,
 		width: 240,
+		minWidth: -Infinity,
+		maxWidth: window.innerWidth,
```
jondlm, line 172: This leaves one final edge case that I can think of. The default `maxWidth` will be pinned to the `maxWidth` when the component mounts (or when the module loads, not sure off the top of my head) which would mean that when the user resizes their browser it wouldn't be taken into account.

Perhaps a better approach would be to set this to `Infinity` and then in the `handleResize` handler check if `this.props.maxWidth === Infinity` and read directly from `window.innerWidth` there. I _think_ that would avoid that edge case.

src/components/SidePanel/SidePanel.tsx
```diff
@@ -172,19 +186,24 @@ class SidePanel extends React.Component<ISidePanelProps, ISidePanelState, {}> {
 
 	handleResize = ({ dX }: { dX: number }): void => {
 		const { startWidth } = this.state;
+		const position = startWidth + dX * (this.props.position === 'right' ? -1 : 1);
+		//const adjustedWidth = !_.isNil(this.props.minWidth) && position < this.props.minWidth ? this.props.minWidth
```
jondlm, line 190: Delete this?

PR #1154, opened 2020-06-18
Bude 121

lancempoe:
I have upgraded "react-transition-group' library as there were many errors and warning being through in our consumers. This library will not work after the next react upgrade as noted in the warnings (deprecated lifecycles). This included the migration as documented here: https://github.com/reactjs/react-transition-group/blob/master/Migration.md

## PR Checklist

Storybook can be viewed [here](https://docspot.adnxs.net/projects/lucid/BUDE-121)

  - [x ] Chrome
  - [ x] Firefox
  - [ x] Safari

- [x ] Unit tests written (`common` at minimum)

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,7 +68,7 @@
 		"react-day-picker": "^6.2.1",
 		"react-motion": "^0.5.2",
 		"react-peek": "^1.0.5",
-		"react-transition-group": "^1.2.1",
+		"react-transition-group": "^4.4.1",
 		"reselect": "^4.0.0"
 	},
 	"devDependencies": {
@@ -100,7 +100,7 @@
 		"@types/react": "^16.8.23",
 		"@types/react-dom": "^16.8.4",
 		"@types/react-motion": "0.0.29",
-		"@types/react-transition-group": "^1.1.6",
+		"@types/react-transition-group": "^4.4.0",
 		"@types/sinon": "^7.0.13",
 		"@typescript-eslint/eslint-plugin": "^1.13.0",
 		"@typescript-eslint/parser": "^1.13.0",
--- a/src/components/Banner/Banner.tsx
+++ b/src/components/Banner/Banner.tsx
@@ -1,7 +1,7 @@
 import _ from 'lodash';
 import React from 'react';
 import PropTypes from 'react-peek/prop-types';
-import { CSSTransition, TransitionGroup } from 'react-transition-group';
+import { CSSTransition } from 'react-transition-group';
 import { lucidClassNames } from '../../util/style-helpers';
 import { omitProps, StandardProps } from '../../util/component-types';
 import CloseIcon from '../Icon/CloseIcon/CloseIcon';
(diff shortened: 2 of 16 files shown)


package.json
```diff
@@ -68,7 +68,7 @@
 		"react-day-picker": "^6.2.1",
 		"react-motion": "^0.5.2",
 		"react-peek": "^1.0.5",
-		"react-transition-group": "^1.2.1",
+		"react-transition-group": "^4.4.1",
```
lancempoe, line 71: This was the the upgrade. All other changes were a result of this migration to version 4.4.1

src/components/Banner/Banner.tsx
```diff
@@ -1,7 +1,7 @@
 import _ from 'lodash';
 import React from 'react';
 import PropTypes from 'react-peek/prop-types';
-import ReactTransitionGroup from 'react-transition-group/CSSTransitionGroup';
+import { CSSTransition, TransitionGroup } from 'react-transition-group';
```
lancempoe, line 4: This was the recommended approach here: https://github.com/reactjs/react-transition-group/blob/master/Migration.md

src/components/Banner/Banner.tsx
```diff
@@ -88,47 +88,48 @@ export const Banner = (props: IBannerProps): React.ReactElement => {
 	}
 
 	return (
-		<ReactTransitionGroup
```
lancempoe, line 91: All I did was replace this component with the two below. The rest of the code is spacing issue.

src/components/Banner/Banner.tsx
```diff
@@ -88,47 +88,48 @@ export const Banner = (props: IBannerProps): React.ReactElement => {
...
+						{isCloseable ? (
+							<CloseIcon
+								isClickable
+								size={8}
+								className={cx('&-close')}
+								onClick={handleClose}
+							/>
+						) : null}
+					</section>
+				) : <React.Fragment></React.Fragment>}
```
lancempoe, line 130: This new component will not accept null so we need to send something, in this case I chose an empty fragment for each case.

src/components/Overlay/Overlay.tsx
```diff
@@ -185,13 +185,14 @@ class Overlay extends React.Component<IOverlayProps, IOverlayState, {}> {
...
-						transitionLeaveTimeout={300}
-					>
-						{overlayElement}
-					</ReactTransitionGroup>
+					<TransitionGroup>
+						<CSSTransition
+							classNames={cx('&')}
+							timeout={300}
+						>
+							{overlayElement || <React.Fragment></React.Fragment>}
```
lancempoe, line 193: The overlayElement might not exist so we need to provide a real react component, so I chose an empty fragment

src/components/Overlay/__snapshots__/Overlay.spec.jsx.snap
```diff
@@ -4,29 +4,31 @@ exports[`Overlay [common] example testing should match snapshot(s) for basic 1`]
...
-    transitionLeaveTimeout={300}
-    transitionName="lucid-Overlay"
-  />
+  <TransitionGroup
+    childFactory={[Function]}
+    component="div"
+  >
+    <CSSTransition
+      classNames="lucid-Overlay"
+      timeout={300}
```
lancempoe, line 13: If timeout is the same for enter and leave, you can now just pass a number for both as shown

src/components/OverlayWrapper/OverlayWrapper.tsx
```diff
@@ -96,23 +96,24 @@ export const OverlayWrapper = (
...
+						<div
+							className={cx('&-message-container', {
+								'&-has-overlay': hasOverlay,
+								'&-kind-light': hasOverlay && overlayKind === 'light',
+								'&-anchored-message': anchorMessage,
+							})}
+						>
+							<div {...messageElementProp} />
+						</div>
+					) : <React.Fragment></React.Fragment>}
```
lancempoe, line 114: Replace the missing child for a fragment as shown here.

src/components/SwitchLabeled/SwitchLabeled.jsx
```diff
@@ -123,19 +123,20 @@ const SwitchLabeled = createClass({
...
+						classNames={cx('&-text')}
+						timeout={100}
+						style={{ position: 'relative' }}
+						className={cx('&-text')}
+					>
+						{labelChildProps ? (
+							<span key={this._labelKey}>
+								{labelChildProps.children || labelChildProps}
+							</span>
+						) : <React.Fragment></React.Fragment>}
```
lancempoe, line 137: Default to empty fragment as shown since a react child is required.

src/components/Table/Table.tsx
```diff
@@ -420,7 +420,7 @@ export class Th extends React.Component<IThProps, IThState> {
 		passiveWidth: this.props.width || null,
 	};
 
-	componentWillReceiveProps({ width }: { width?: number | string | null }) {
+	UNSAFE_componentWillReceiveProps({ width }: { width?: number | string | null }) {
```
lancempoe, line 423: This was also showing up in our consumers so I fixed the last one in lucid.

PR #1159, opened 2020-07-28
1155 add draggable linechart

AndreEPaul:
## PR Checklist

Storybook can be viewed [here](https://docspot.adnxs.net/projects/lucid/1155-add-draggable-linechart/?path=/docs/documentation-introduction--introduction)

  - [x] Chrome
  - [x] Firefox
  - [x] Safari


package.json
```diff
@@ -52,14 +52,24 @@
...
 		"@babel/parser": "^7.2.2",
+		"@types/d3-array": "^2.0.0",
+		"@types/d3-axis": "^1.0.12",
+		"@types/d3-drag": "^1.2.3",
+		"@types/d3-selection": "^1.4.2",
+		"@types/d3-transition": "^1.1.6",
 		"@types/hoist-non-react-statics": "^3.3.1",
 		"classnames": "^2.2.3",
 		"create-react-class": "^15.6.2",
+		"d3-array": "^2.4.0",
```
lancempoe, line 63: Is there any reason we have all of these d3 packages. How about just importing d3 (npm install d3). I believe you can still just import in in the file things like d3-axis. https://www.npmjs.com/package/d3

package.json
```diff
@@ -52,14 +52,24 @@
 	"license": "Apache-2.0",
 	"dependencies": {
 		"@babel/parser": "^7.2.2",
+		"@types/d3-array": "^2.0.0",
```
lancempoe, line 55: My question is similar here: npm install --save @types/d3. This might be easier to maintain as we only update 1 file in package instead of many

src/components/DraggableLineChart/DraggableLineChart.tsx
```diff
@@ -0,0 +1,170 @@
...
+} from '../../util/component-types';
+import * as d3Selection from 'd3-selection';
+import { lucidClassNames } from '../../util/style-helpers';
+import DraggableLineChartD3 from './DraggableLineChartD3';
+
+const cx = lucidClassNames.bind('&-DraggableLineChart');
+const { arrayOf, number, object, shape, func } = PropTypes;
+
+interface IDraggableLineChartMargin {
+	top: number;
```
lancempoe, line 17: wow, tabs?

src/components/DraggableLineChart/DraggableLineChart.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+	/**
+	 * Data for the chart. E.g.
+	 * { x: new Date('2015-01-01') , y: 1 } ,
+	 * { x: new Date('2015-01-02') , y: 2 } ,
+	 * { x: new Date('2015-01-03') , y: 3 } ,
+	 * { x: new Date('2015-01-04') , y: 2 } ,
+	 * { x: new Date('2015-01-05') , y: 5 } ,
+	 * ]
+	 * (brought in from LineChart)
```
lancempoe, line 44: I wouldn't keep this comment as LineChart can change separate from this work.

src/components/DraggableLineChart/DraggableLineChart.tsx
```diff
@@ -0,0 +1,170 @@
...
+	 */
+	data: Array<{ [key: string]: Date | string | number | undefined }>;
+
+	/**
+	 * Drag handler function which is a callable function executed at the end of drag
+	 */
+	onDragEnd?: (d: any) => any;
```
lancempoe, line 51: any chance we can replace any with anything else?

src/components/DraggableLineChart/DraggableLineChart.tsx
```diff
@@ -0,0 +1,170 @@
...
+}
+
+export type IDraggableLineChartProps = Overwrite<
+	React.SVGProps<SVGGElement>,
+	IDraggableLineChartPropsRaw
+>;
+
+class DraggableLineChart extends React.Component<IDraggableLineChartProps, {}> {
+	ref: any;
```
lancempoe, line 60: same here, we really should avoid any like the plague.

src/components/DraggableLineChart/DraggableLineChartD3.ts
```diff
@@ -0,0 +1,186 @@
...
+			bottom: 65,
+			left: 80,
+		},
+		height: 300,
+		width: 1000,
+		onDragEnd: _.noop,
+		data: [],
+		cx: _.noop,
+	};
+	xScale: any;
```
lancempoe, line 39: There is a lot of use of any. Let's either remove this or turn to jsx. TSX is just interfaces, so if you use any, we are not using the power of tsx

src/components/DraggableLineChart/DraggableLineChartD3.ts
```diff
@@ -0,0 +1,186 @@
...
+			.duration(500)
+			.attr('cx', (d: any) => this.xScale(d.x))
+			.attr('cy', (d: any) => this.yScale(d.y))
+			.attr('r', 5)
+			.style('fill', '#009fdb')
+			.style('stroke', 'white')
+			.style('stroke-width', 1);
+		if (isNew) circle.call(this.drag());
+	};
+	renderLineChart = () => {
```
lancempoe, line 169: any reason we don't separate all of these 'renders' to their own react component? Was it issues with lifecycles? It feels a little code by config, which is hard to maintain.

package.json
AndreEPaul: That makes sense, I think the initial idea was that is how it was being done here (`package.json` was already including some d3 packages but not all) probably to prevent including things we don't need. But yeah with this component specifically we are using more d3 packages. Changing this however would require changes to anything in lucid which imports a d3 package (change being `import ... from 'd3-specificPackage'` -> `import ... from 'd3'`)

src/components/DraggableLineChart/DraggableLineChart.tsx
AndreEPaul: Hm weird. Shouldn't this be addressed by `prettier`?

lancempoe: yes but It might not have ran on commit

package.json
yasskin: It seems okay with me to follow the pattern that's there of importing individual packages. We could do a separate PR for d3.

jhsu: > Scope-wise this feels like a separate PR if we would like to only import d3 as a whole.
agreed

also, importing separate d3 packages makes sense to me since we're clear about the dependency and not importing an umbrella package, though it makes it easier, it isn't clear about what part we're using and has more than we need. we don't use the dom part of d3, I don't think we should pull in d3.

src/components/DraggableLineChart/DraggableLineChart.tsx
AndreEPaul: Weird. Ran it locally to be sure and it shows no diff with this branch but it looks like no tabs on my end

src/components/DraggableLineChart/DraggableLineChartD3.ts
yasskin: Agree that 'any' kinda defeats the purpose of TypeScript but we've had difficulties making d3 play nicely with TS. Take a look at Lines.tsx and Points.tsx for some examples of how we handled xScale and yScale with TS. You can probably explicitly identify a type for these scales.

src/components/DraggableLineChart/DraggableLineChart.tsx
lancempoe: prettier might not be setup on this repo.

PR #1226, opened 2021-04-07
Bude 1069 1 add truncating ellipses to header columns

NYConfidential:
## PR Checklist

Storybook can be viewed [here](https://docspot.adnxs.net/projects/lucid/BUDE-1069_1_add_truncating_ellipses_to_header_columns)

  - [ x] Chrome
  - [] Firefox
  - [x ] Safari

- [x ] Unit tests written (`common` at minimum)
- [x ] PR has one of the `semver-` labels
- [ ] 
image

### DIFF
--- a/src/components/DataTable/DataTable.tsx
+++ b/src/components/DataTable/DataTable.tsx
@@ -22,22 +22,14 @@ const cx = lucidClassNames.bind('&-DataTable');
 const cxe = lucidClassNames.bind('&-DataTable-EmptyStateWrapper');
 const SELECTOR_COLUMN_WIDTH = 41;
 
-const {
-	any,
-	func,
-	number,
-	object,
-	string: stringProps,
-	bool,
-	arrayOf,
-} = PropTypes;
+const { any, func, number, object, string: stringProps, bool, arrayOf } = PropTypes;
 
-interface IDataTableProps
-	extends ITableProps,
-		React.DetailedHTMLProps<
-			React.HTMLAttributes<HTMLDivElement>,
-			HTMLDivElement
-		> {
+interface IDataTableProps extends ITableProps,
+	React.DetailedHTMLProps<
+	React.HTMLAttributes<HTMLDivElement>,
+	HTMLDivElement
+> {
+
 	/*
 	 * 	Array of objects to be rendered in the table. Object keys match the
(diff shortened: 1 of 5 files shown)


src/components/DataTable/DataTable.tsx
```diff
@@ -428,14 +439,14 @@ export const DataTable = (props: IDataTableProps) => {
 													!_.isNil(columnProps.hasBorderRight)
 														? columnProps.hasBorderRight
 														: isFixedColumn &&
-														columnIndex + 1 + (isSelectable ? 1 : 0) ===
-														endColumn
+														  columnIndex + 1 + (isSelectable ? 1 : 0) ===
+																endColumn
```
scotthollandjr, line 443: i'm worried about this indentation. i assume it was auto-indented but still looks a little weird

src/components/Table/Table.less
```diff
@@ -209,22 +209,42 @@
 				}
 			}
 
+
+			// header cell content displays ellipses if text overflows
+			// and <DataTable hasFixedHeader={true} truncateContent={true}>
+			&.@{prefix}-Table-truncate-content {
+				& > .@{prefix}-Table-Th-inner {
+					//display: flex;
```
scotthollandjr, line 217: unnecessary comment

scotthollandjr (approved):

ginoafable-xandr (approved):
Approved - Pending: change the word "Occupation" to something like "Daytime Occupation" or make the column narrower to show "Occupati..." to show a better example. Other users might not know why the elipses is there after the whole word was shown if they don't open the code and find out the column is actually "Occupation Occupation Occupation" which is not a common header title.
Screen Shot 2021-04-08 at 3 01 50 PM

src/components/DataTable/DataTable.tsx
NYConfidential: Yes, it's the linter... But should be fine

lancempoe (changes requested):
Please keep all PRs only related to the task at hand. There are all kinds of changes here that make it really hard to review. 

If there was a linting change recently, then these files need updated PRIOR to this change.

lancempoe (approved):

PR #1250, opened 2021-08-30
refactor: update A icons to SB6 formt

lennoxdavidjr:
## PR Checklist

  - [ ] Chrome
  - [ ] Firefox
  - [ ] Safari


yasskin (approved):
Great upgrades to the SB6 functionality. One observation: we might want to give some thought to a future project where we reorder the props systematically, so the interactive ones show up closer to the top. On DoubleChevronIcon, the direction props are at the bottom, so it's hard to see the effect on the icon when you change the direction. 
Screen Shot 2021-08-31 at 9 52 28 AM

lennoxdavidjr:
Excellent suggestion @yasskin I can implement that in the next PR, should be very easy to do 👍🏿

dmaloney01 (approved):
UX approved. Ideally I don't want to expose the direction prop on icons where it doesn't apply. There are only a few icons (typically arrows and chevrons) where the direction comes into play. We don't want to encourage the left/right flipping of the other icons.

lennoxdavidjr:
@dmaloney01 Understood. I'll update those. Thanks for the review

jhaage (approved):

PR #1251, opened 2021-08-31
refactor: update backuparrowicon, pinicon and props order

lennoxdavidjr:
## PR Checklist

Storybook can be viewed [here](https://docspot.adnxs.net/projects/lucid/CXP-2164-Update-Icons-SB6-Part-2/?path=/docs/documentation-intro--page)

  - [x] Chrome
  - [x] Firefox
  - [x] Safari

* PR updates "B" and "C" icons to Storybook 6 compatible format. This includes interactive controls via the UI


jhaage (approved):

dmaloney01 (approved):
UX approved

dreidn (approved):

yasskin (changes requested):
Looks like the Bell icon dots aren't working. It would be nice to keep the sample stories too, with examples of each color. See the old example:
https://docspot.adnxs.net/projects/lucid/CXP-2164-Update-Icons-SB6/?path=/docs/icons-icons-bellicon-sb5--basic
Clicking on featured color isn't setting the dot currently:
https://docspot.adnxs.net/projects/lucid/CXP-2164-Update-Icons-SB6-Part-2/?path=/docs/icons-icons-bellicon--primary

yasskin (changes requested):
Looks like the CloseIcon colors aren't changing. This Icon is a bit different because the default is a grey not our primary blue. I wonder if it should have the color options at all ? And if Primary for it is grey, not blue...
https://docspot.adnxs.net/projects/lucid/CXP-2164-Update-Icons-SB6-Part-2/?path=/docs/icons-icons-closeicon--primary

lennoxdavidjr:
Very good eye @yasskin If I understand correctly, Close Icon is a special case with special rules governing its color. Looking at the story, it appears that the primary and only color for Close Icon is grey. This icon also has a special feature that changes color slightly when the user hovers over the icon and the `isClickable` prop is set to true and the `isDisabled` prop is set to false. I agree that in this instance that the color feature should be disabled. I will remove that now

lennoxdavidjr:
Additionally, @yasskin I am having no issues changing the bell icon colors or dot colors. I am happy to pair to see if this is an issue

PR #1259, opened 2021-09-14
Cxp 2042 update selected state UI

yasskin:
## PR Checklist

Storybook for ButtonGroup can be viewed [here](https://docspot.adnxs.net/projects/lucid/cxp-2042-update-selected-state-ui/?path=/docs/controls-buttongroup--default)
Storybook for SingleSelect can be viewed [here](https://docspot.adnxs.net/projects/lucid/cxp-2042-update-selected-state-ui/?path=/docs/controls-singleselect--default)
Storybook for Searchable Select can be viewed [here](https://docspot.adnxs.net/projects/lucid/cxp-2042-update-selected-state-ui/?path=/docs/controls-searchableselect--default)

This PR also corrects a few typos and standardizes some css variables.

  - [x] Chrome
  - [ ] Firefox
  - [ ] Safari


jhaage:
While interacting with the Storybook for ButtonGroup I found the behavior of de-selecting a button to be strange. It's not something that was introduced in this PR, I see the existing ButtonGroup has the same behavior, but it seems misleading. It's easiest to see in the **Stateless** example:

1. Hover over Zero and see the gray background
2. Click Zero and the blue background and border appear
3. Click Zero again to de-select that button
4. Even when not hovered, the button remains with a gray background which makes it seem selected
5. If you now click on another button, the gray background on the Zero button will disappear

Behavior is the same prior to these recent changes, so it was not just introduced.

@dmaloney01 curious on your thoughts if you'd also consider this behavior a bug or if it's working as expected?

jhaage:
On the SearchableSelect component, I noticed in the **loading** story that it selects the wrong index. This again is something I can reproduce in the existing Lucid component, so it wasn't just introduced. All the other examples seem to select the correct index.
https://www.screencast.com/t/QVsS52txUoig

jhaage (approved):
I made a couple comments on some behavior that seemed unexpected to me, but they are both existing behavior (not newly introduced). These changes look good to me.

jhaage:
> While interacting with the Storybook for ButtonGroup I found the behavior of de-selecting a button to be strange. It's not something that was introduced in this PR, I see the existing ButtonGroup has the same behavior, but it seems misleading. It's easiest to see in the **Stateless** example:
> 
>     1. Hover over Zero and see the gray background
> 
>     2. Click Zero and the blue background and border appear
> 
>     3. Click Zero again to de-select that button
> 
>     4. Even when not hovered, the button remains with a gray background which makes it seem selected
> 
>     5. If you now click on another button, the gray background on the Zero button will disappear
> 
> 
> Behavior is the same prior to these recent changes, so it was not just introduced.
> 
> @dmaloney01 curious on your thoughts if you'd also consider this behavior a bug or if it's working as expected?

I have a feeling this is because there is both **hovered** and **focused** and they both have the gray color. When you de-select and are not hovering, the button must still be **focused**?

.storybook/main.js
```diff
@@ -1,4 +1,4 @@
-export default {
```
yasskin, line 1: When building Storybook, I get this warning, which goes away when the file is converted to a .js file.
```
info => Adding stories defined in "/Users/nyasskin/Documents/lucid/.storybook/main.js"
info => Loading custom manager config
WARN Unable to find main.js addons: /Users/nyasskin/Documents/lucid/.storybook/main
```

velhala (approved):
Let's check with @tsantef  about the main.js (.ts) issue with the storybook building.

dmaloney01 (approved):

lennoxdavidjr (approved):

lennoxdavidjr:
@yasskin you may need to rebase to pull in Tim's render changes

.storybook/main.js
yasskin: I reverted this to main.ts -- still get the warning but storybook works.

PR #1270, opened 2021-10-12
BUDE-1535: Add `special` prop to `TextFieldValidated` that allows custom error text and border color

scotthollandjr:
## PR Checklist

Update `TextFieldValidated` to accept a `special` prop with text and border color strings to style informational messages

  - [x] Chrome
  - [x] Firefox
  - [x] Safari

Screen Shot 2021-10-12 at 4 48 51 PM


scotthollandjr:
> Do we need the blue "Info" style version for a particular use case? My concern with it is we often show selected states and active states in our components with a blue outline. This might cause some confusion.

Yes, this was requested by product and actually exists already as the `Info` prop. now we're adding a green variant which is why I want to make it more flexible for future iterations

n8thegr8 (approved):
LGTM, after the conflicts are resolved.

scotthollandjr:
Out of date, re-submitting [here](https://github.com/appnexus/lucid/pull/1352)

PR #1280, opened 2021-11-10
Add propTypes to Banner component

yasskin:
## PR Checklist

Storybook can be viewed [here](https://docspot.adnxs.net/projects/lucid/banner-fix)

Add Common test for propTypes for components
Fix propTypes on Paginator story and component
Fix Storybook Categories
Add FC type defs to Badge and Banner
Remove invalid undefined from Tabs
Add new properties to FC interface
Change ExampleComponent to Functional Component
Add missing args to Banner stories

  - [x] Chrome
  - [ ] Firefox
  - [ ] Safari

### DIFF
--- a/src/components/Badge/Badge.tsx
+++ b/src/components/Badge/Badge.tsx
@@ -2,7 +2,7 @@ import _ from 'lodash';
 import React from 'react';
 import PropTypes from 'prop-types';
 import { lucidClassNames } from '../../util/style-helpers';
-import { omitProps, StandardProps } from '../../util/component-types';
+import { omitProps, FC, StandardProps } from '../../util/component-types';
 
 const cx = lucidClassNames.bind('&-Badge');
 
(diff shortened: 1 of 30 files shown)


src/components/Badge/Badge.tsx
```diff
@@ -59,7 +60,9 @@ const defaultProps = {
 	type: Type.filled,
 };
 
-export const Badge = (props: IBadgeProps): React.ReactElement => {
+export const Badge: FC<IBadgeProps> = (
+	props: IBadgeProps
+): React.ReactElement => {
 	const { className, kind, type, children, ...passThroughs } = props;
```
rcousineau-xandr, line 66: I've seen this written as...

```ts
export const Badge: FC<IBadgeProps> = ({ className, kind, type, children, ...passThroughs }) => {
```

which is basically the same but kinda looks cleaner in my opinion. 

No need to change it, but I thought I'd mention in case that looks appealing to you 😄

yasskin: Thanks for the suggestion. I think that's easier to read. Maybe it makes sense to do that rather than destructure the props on line 66.

PR #1331, opened 2022-02-03
OE-975 : add data-test-id to lucid portals

swetanshkumar:
## PR Checklist

This is about adding data-test-ids to portal so that we can use them in our cypress tests to verify the UI behaviour around the dialog and other UI Elements where portal is being used. 
I added className as data-test-id because it seems to be the best way, as if I used portalId, it was something autogenerated one which we can't use to test the behaviour.

  - [x] Chrome
  - [x] Firefox
  - [x] Safari

### DIFF
--- a/src/components/Portal/Portal.spec.tsx
+++ b/src/components/Portal/Portal.spec.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import assert from 'assert';
-import { mount } from 'enzyme';
+import { mount, shallow } from 'enzyme';
 
 import Portal from './Portal';
 
@@ -12,6 +12,15 @@ describe('Portal', () => {
 
 				assert(document.getElementById('test1234'));
 				wrapper.unmount();
+			});
+		});
+
+		describe('data-test-id', () => {
+			it('should set the data-test-id of the portal DOM element portalId', () => {
+				const wrapper = shallow(<Portal className='test-classname' />);
+				expect(
+					wrapper.find('[data-test-id="test-classname"]').exists()
+				).toEqual(true);
 			});
 		});
 
--- a/src/components/Portal/Portal.tsx
+++ b/src/components/Portal/Portal.tsx
@@ -78,6 +78,7 @@ class Portal extends React.Component<IPortalProps, IPortalState, {}> {
 		return this.state.isReady
 			? ReactDOM.createPortal(
 					<div
+						data-test-id={this.props.className}
 						className={classNames(cx('&'), this.props.className)}
 						{...omitProps(this.props, undefined, _.keys(Portal.propTypes))}
 					>


src/components/Portal/Portal.spec.tsx
```diff
@@ -15,6 +15,15 @@ describe('Portal', () => {
 			});
 		});
 
+		describe('data-test-id', () => {
```
n8thegr8, line 18: This should be `data-testid`

src/components/Portal/Portal.spec.tsx
```diff
@@ -15,6 +15,15 @@ describe('Portal', () => {
...
+			it('should set the data-test-id of the portal DOM element portalId', () => {
+				const wrapper = shallow(<Portal className='test-classname' />);
+				expect(
+					wrapper.find('[data-test-id="test-classname"]').exists()
```
n8thegr8, line 22: Same as above `data-testid`

Also, I'm a little confused why the className is being passed to the data-testid.  We should just create a new prop name dataTestid and pass that down to the portal.

n8thegr8 (changes requested):
A few small changes.

src/components/Portal/Portal.tsx
```diff
@@ -78,6 +78,7 @@ class Portal extends React.Component<IPortalProps, IPortalState, {}> {
 		return this.state.isReady
 			? ReactDOM.createPortal(
 					<div
+						data-test-id={this.props.className}
```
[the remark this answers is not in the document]
swetanshkumar: We use `data-test-id` for our cypress selectors. Please refer to the slack thread where I answered it.

src/components/Portal/Portal.spec.tsx
swetanshkumar: Same here as well

src/components/Portal/Portal.spec.tsx
swetanshkumar: I do use the className and not separate props passing down the component because this change will just add our `data-test-id` to elements without doing additional changes to the parent components calling this.

If we use a separate prop, it will have affects on other parent component calling this and change will be large enough which seems to me an additional effort for this.

semaurer: @swetanshkumar I think it makes more sense to follow @n8thegr8's suggestion.  This is an open source library and I think consumers would expect that you could pass down a data-testid prop that would get added to the component with the same prop name.  I don't think using the classname for the data-testid prop would be expected.

swetanshkumar: @semaurer , then this might more changes to the parent component used. Ideally we don't directly call lucid portals in our apps, its generally the parent components which passed down the props and portal is being rendered.

For example - 
Take Dialog as an example, We generally use that pretty much. So the flow is - 

Dialog ---> Overlay ---> Portals

if we introduce a separate prop, we might have to pass it from Dialog as well. So its like you have to update Dialog to accept data-test-id as a prop to pass down to Overlay, then add data-test-id to Overlay as a prop to accept and pass it along Portals


Note - This is just an example of Dialog. There are still many (SearchableMultiSelect ---> DropMenu ---> ContextMenu ---> Portals)

src/components/Portal/Portal.tsx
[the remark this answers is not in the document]
n8thegr8: Ahh, we should probably change that over to data-testid eventually since that seems to be best practice and it is natively used by Mabl.

src/components/Portal/Portal.spec.tsx
n8thegr8: data-test-ids should be unique whereas class names are not.

src/components/Portal/Portal.tsx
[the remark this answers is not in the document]
nathaniellee: We use `data-test-id` rather than `data-testid` throughout our internal applications, which is why you'll see it used within Lucid. Migrating to a new naming convention is a bigger conversation for the team.

src/components/Portal/Portal.spec.tsx
nathaniellee: We currently do not use `data-test-id` props much at all within Lucid — you'll find them only on the "Skeleton" family of components at the moment, and they are certainly not unique — so there is an established precedent within the library.

I agree with @swetanshkumar that @n8thegr8's suggestion is a bigger discussion that we need to have insofar as how we're going to handle this prop across the board in a consistent manner for all of the published components. If we cover that in this particular PR, then as @swetanshkumar said we would need to update not just the Portal component but every component within Lucid that consumes that component.

An interim solution would be to set a default value for the `data-test-id` prop that utilizes some sort of unique ID generator in order to ensure uniqueness, but I strongly feel that we should look to handling this uniformly across the component library through a deliberate, concerted effort rather than a temporary, one-off solution.

nathaniellee (approved):

n8thegr8 (approved):

src/components/Portal/Portal.tsx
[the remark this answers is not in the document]
n8thegr8: Yes, I agree we should definitely have a larger discussion for data-test-id vs data-testid.  Sorry, I didn't mean we should do it now.

PR #1391, opened 2022-07-26
feat: add onClick handler to Tag component

lennoxdavidjr:
## PR Checklist

Description of changes:

Link(s) to [Storybook](https://docspot.adnxs.net/projects/lucid/CXP-3283-Add-OnClick-Tag/?path=/docs/communication-tag--basic) on docspot:

  - [x] Chrome
  - [x] Firefox
  - [x] Safari

- [ ] Two cxp team engineer approvals
- [ ] One product design approval (as necessary)

### DIFF
--- a/src/components/Tag/Tag.tsx
+++ b/src/components/Tag/Tag.tsx
@@ -66,7 +66,6 @@ const defaultProps = {
 	isClickable: false,
 	isRemovable: false,
 	kind: 'default' as const,
-	onClick: _.noop,
 	onRemove: _.noop,
 };
 
@@ -90,7 +89,7 @@ export const Tag = (props: ITagProps): React.ReactElement => {
 
 	const handleClick = (event: React.MouseEvent) => {
 		if (isClickable && onClick) {
-			return onClick({ event, props: props });
+			return onClick({ event, props });
 		}
 
 		return;
--- a/src/components/Tag/Tag.tsx
+++ b/src/components/Tag/Tag.tsx
@@ -39,7 +39,7 @@ export interface ITagPropsRaw extends StandardProps {
 	/** Style variations of the `Tag`. */
 	kind?: 'primary' | 'success' | 'warning' | 'danger' | 'info' | 'default';
 
-	onClick: ({
+	onClick?: ({
 		event,
 		props,
 	}: {
@@ -89,7 +89,7 @@ export const Tag = (props: ITagProps): React.ReactElement => {
 	};
 
 	const handleClick = (event: React.MouseEvent) => {
-		if (isClickable) {
+		if (isClickable && onClick) {
 			return onClick({ event, props: props });
 		}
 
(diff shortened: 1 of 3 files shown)


src/components/Tag/Tag.tsx
```diff
@@ -29,12 +30,23 @@ export interface ITagProps
 		background (e.g. in a page header). */
 	hasLightBackground: boolean;
 
+	/** Allows Tag to be interacted with via the onClick function */
+	isClickable?: boolean;
```
tsantef, line 34: I believe we normally use the existence of the onClick prop to mean the component is clickable. @yasskin can you confirm this?

tsantef (changes requested):

dreidn (approved):

src/components/Tag/Tag.tsx
dreidn: We already use an isClickable prop for icons, but the existence of onClick seems like it would work. If isClickable is true, there should always be an onClick

src/components/Tag/Tag.tsx
```diff
@@ -99,7 +122,7 @@ export const Tag = (props: ITagProps): React.ReactElement => {
 		>
 			{hasOtherChildren && (
 				<span className={cx('&-other-children')}>
-					{otherChildren}
+					<span onClick={handleClick}>{otherChildren}</span>
 					{isRemovable && (
 						<CloseIcon
 							onClick={handleRemove}
```
yasskin, line 128: I think the design of this component is to handle the onClick event just with the CloseIcon. So you set isRemovable to true and handle the click event in the "handleRemove" handler.

src/components/Tag/Tag.tsx
yasskin: I think this works. It makes the entire tag clickable. I just wonder if the intent of the design of the component is to only have the CloseIcon clickable, not the entire tag.
Take a look at the Interactive story: https://docspot.adnxs.net/projects/anx-react/latest/?path=/docs/ludic-design-system_communication-tag--interactive

yasskin: > I believe we normally use the existence of the onClick prop to mean the component is clickable. @yasskin can you confirm this?

The `isClickable` prop and the pattern used in this PR is used all over the place in the repo, in Icons, for example. @tsantef

dreidn (changes requested):
Can you add a story to show to to use onClick and isClickable? 
If both the onClick for the span and onClick for CloseIcon are provided, and you click the close icon, does it fire both event handlers? You might need to stop the click event from the CloseIcon from bubbling up, otherwise consuming apps will need to handle it.

https://stackoverflow.com/questions/38619981/how-can-i-prevent-event-bubbling-in-nested-react-components-on-click

yasskin (commented):
If we go with the `isClickable` route, I'd like to see a unit test added for that prop.

src/components/Tag/Tag.tsx
dreidn: @yasskin I think that was the original design of the component, but I'm guessing @lennoxdavidjr needs the whole tag to be clickable for something related to the global search update? There should be a story showing how the props are used though

lennoxdavidjr:
Thanks for the feedback everyone. On second consideration, I think that this change may not be necessary at this moment so I am going to close this PR. Should this need arise again, we can revisit

